PR #2, opened 2021-07-14
Add config_settings to echo config details in init span

johnbley:


flands (approved):

johnbley:
Starting over after getting signed commits working

PR #50, opened 2022-02-12
screen render time

Piyush23786:
Capture screen render time events enabling detection of slow/frozen frames


seemk (commented):
Is some code missing here to create the spans?

PR #51, opened 2022-02-14
Feature/apmi 2307 session timeout 

shattimare:
Implemented `15 minutes `session time out logic with existing ` 4 hours `session time out logic.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
@@ -19,6 +19,9 @@ import Foundation
 import UIKit
 import OpenTelemetrySdk
 
+let INACTIVITY_SESSION_TIMEOUT_SECONDS = 15 * 60
+private var sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+
 func initializeAppLifecycleInstrumentation() {
     /*
      Observed event sequences:
@@ -57,6 +60,7 @@ func initializeAppLifecycleInstrumentation() {
 }
 var activeSpan: SpanHolder?
 func lifecycleEvent(_ event: String) {
+    invalidateSession(event)
     // these two start spans
     if event == "UIApplicationWillResignActiveNotification" ||
             event == "UIApplicationWillEnterForegroundNotification" {
@@ -92,5 +96,15 @@ func lifecycleEvent(_ event: String) {
             event == "UIApplicationDidEnterBackgroundNotification" {
         OpenTelemetrySDK.instance.tracerProvider.forceFlush()
 
+    }
+}
+func invalidateSession(_ event: String) {
+    // 15 min inactivity then session time out
+    if event == "UIApplicationWillResignActiveNotification" {
+        sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+    } else if event == "UIApplicationWillEnterForegroundNotification" {
+        if Date() > sessionIdInActivityExpiration { // expire 15 min
+            rumSessionId = generateNewSessionId()
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -52,16 +54,22 @@ func getRumSessionId() -> String {
             sessionIdLock.unlock()
         }
     }
-    if Date() > sessionIdExpiration {
+    if Date() > sessionIdExpiration { // expire in 4 hours
         sessionIdExpiration = Date().addingTimeInterval(TimeInterval(MAX_SESSION_AGE_SECONDS))
+        sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(MAXSESSIONINACTIVEAGESECONDS))
         rumSessionId = generateNewSessionId()
         callbacks = sessionIdCallbacks
+    } else if Date() > sessionIdInActivityExpiration { // expire 15 min
```
johnbley, line 62: There's nothing in here actually checking for inactivity.  This check should be done when the app is being brought back to the foreground, having been in the background for more than the `INACTIVE` timeout.  Such logic probably belongs in `AppLifecycle.swift` rather than here, with a method like `invalidateSession` placed here that the inactivity logic can call.

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -18,9 +18,11 @@ limitations under the License.
 import Foundation
 
 let MAX_SESSION_AGE_SECONDS = 4 * 60 * 60
+let MAXSESSIONINACTIVEAGESECONDS = 15 * 60
```
johnbley, line 21: Snake-case this and name it appropriately please.  `INACTIVITY_SESSION_TIMEOUT_SECONDS`

shattimare:
> Logic doesn't actually check for inactivity.

Implemented suggested changes to detect inactivity in AppLifeCycle.swift.

Previous logic was implemented in that manner because we got confused with below line mentioned in JIRA ticket:
"User inactivity is defined as a 15 minute period during which **our instrumentations** have not detected any user interactions with the app( please verify if the definition matches your expectations)."

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -19,7 +19,7 @@ import Foundation
 
 let MAX_SESSION_AGE_SECONDS = 4 * 60 * 60
 
-private var rumSessionId = generateNewSessionId()
+var rumSessionId = generateNewSessionId()
```
johnbley, line 22: Please don't make this non-private.  Access to it is controlled by the `sessionIdLock` which you have bypassed in your  `invalidateSession`.

johnbley (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
```diff
@@ -19,6 +19,10 @@ import Foundation
 import UIKit
 import OpenTelemetrySdk
 
+let INACTIVITY_SESSION_TIMEOUT_SECONDS = 15 * 60
+private var sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+var Is_New_SessionID_For_In_Activity = false
```
seemk, line 24: Can we avoid random naked globals perhaps? Also the naming/casing should be consistent with other variables. P.S.: `inactivity`

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
```diff
@@ -94,3 +99,14 @@ func lifecycleEvent(_ event: String) {
...
     }
 }
+func invalidateSession(_ event: String) {
+    // 15 min inactivity then session time out
+    if event == "UIApplicationWillResignActiveNotification" {
+        sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+    } else if event == "UIApplicationWillEnterForegroundNotification" {
+        if Date() > sessionIdInActivityExpiration { // expire 15 min
+           Is_New_SessionID_For_In_Activity = true
+            _  = getRumSessionId()
```
seemk, line 109: Can't it just trigger new session id generation here instead?

shattimare: yes, it will create new session id after 15 mins of  inactivity.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
shattimare: Yes,I will change name of `Is_New_SessionID_For_In_Activity` with `Is_New_SessionID_For_InActivity`.
Is_New_SessionID_For_In_Activity variable used to communicate between two files AppLifeCycle and Session. that's  why I make it global.

PR #52, opened 2022-02-14
Feature/apmi 2310 capture session change event

shattimare:
Implemented Instrumentation of session change event.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
@@ -69,7 +69,7 @@ func getRumSessionId() -> String {
         callback()
     }
     if isSessionIdChanged {
-        createSessionIdChangeSpan()
+       // createSessionIdChangeSpan()
     }
     return rumSessionId
 }
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
@@ -17,13 +17,14 @@ limitations under the License.
 
 import Foundation
 
-let MAX_SESSION_AGE_SECONDS = 4 * 60 * 60
+let MAX_SESSION_AGE_SECONDS = 40   // 4 * 60 * 60
 
 private var rumSessionId = generateNewSessionId()
 private var sessionIdExpiration = Date().addingTimeInterval(TimeInterval(MAX_SESSION_AGE_SECONDS))
 private let sessionIdLock = NSLock()
 private var sessionIdCallbacks: [(() -> Void)] = []
 private var isSessionIdChanged = false
+private var oldRumSessionId = ""
 
 func generateNewSessionId() -> String {
     var i=0
@@ -33,7 +34,9 @@ func generateNewSessionId() -> String {
         let b = Int.random(in: 0..<256)
         answer += String(format: "%02x", b)
     }
-    isSessionIdChanged = true
+    if oldRumSessionId != "" { // firtst time id generated so it is not consider as session id changed
+        isSessionIdChanged = true
+    }
     return answer
 }
 
@@ -56,6 +59,7 @@ func getRumSessionId() -> String {
     }
     if Date() > sessionIdExpiration {
         sessionIdExpiration = Date().addingTimeInterval(TimeInterval(MAX_SESSION_AGE_SECONDS))
+        oldRumSessionId = rumSessionId
         rumSessionId = generateNewSessionId()
         callbacks = sessionIdCallbacks
     }
@@ -72,7 +76,7 @@ func getRumSessionId() -> String {
 func createSessionIdChangeSpan(){
     isSessionIdChanged = false
     let tracer = buildTracer()
-    let span = tracer.spanBuilder(spanName: "SessionID Change").setSpanKind(spanKind: .client).startSpan()
-    span.setAttribute(key: "startTime", value: "Date()")
+    let span = tracer.spanBuilder(spanName: "sessionId.change").setSpanKind(spanKind: .client).startSpan()
+    span.setAttribute(key: "splunk.rum.previous_session_id", value: oldRumSessionId)
     span.end()
 }
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
@@ -17,7 +17,7 @@ limitations under the License.
 
 import Foundation
 
-let MAX_SESSION_AGE_SECONDS = 30 //4 * 60 * 60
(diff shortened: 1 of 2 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -17,12 +17,14 @@ limitations under the License.
 
 import Foundation
 
-let MAX_SESSION_AGE_SECONDS = 4 * 60 * 60
+let MAX_SESSION_AGE_SECONDS = 40   // 4 * 60 * 60
```
seemk, line 20: Think this is left in from debugging (40s)

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -62,6 +68,16 @@ func getRumSessionId() -> String {
...
-
+    if isSessionIdChanged {
+        createSessionIdChangeSpan()
+    }
     return rumSessionId
 }
+func createSessionIdChangeSpan(){
+    isSessionIdChanged = false
+    let tracer = buildTracer()
+    let span = tracer.spanBuilder(spanName: "sessionId.change").setSpanKind(spanKind: .client).startSpan()
```
seemk, line 79: No need to specify `client` span here

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -62,6 +68,16 @@ func getRumSessionId() -> String {
...
+    span.setAttribute(key: "splunk.rum.previous_session_id", value: oldRumSessionId)
+    // span.setAttribute(key: "component", value: "appstart")
```
seemk, line 81: Can get rid of this comment

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -62,6 +68,16 @@ func getRumSessionId() -> String {
     for callback in callbacks {
         callback()
     }
-
+    if isSessionIdChanged {
+        createSessionIdChangeSpan()
```
seemk, line 72: Could we invoke this right when expiration is checked? Can get rid of `isSessionIdChanged` then

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
shattimare: changed.

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
shattimare: Removed it.

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
shattimare: remove this line

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
shattimare: As per our discussion, we keep it as it is.

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -62,6 +68,15 @@ func getRumSessionId() -> String {
...
+    }
     return rumSessionId
 }
+func createSessionIdChangeSpan() {
+    isSessionIdChanged = false
+    let tracer = buildTracer()
+    let span = tracer.spanBuilder(spanName: "sessionId.change").startSpan()
```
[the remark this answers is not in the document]
shattimare: i add start time and and time.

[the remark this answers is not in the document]
shattimare: `isSessionIdChanged` is not functional - local variable and it is not static as well, it is used by more than one function of the same swift file.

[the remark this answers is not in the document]
johnbley: > `isSessionIdChanged` is not function - local variable and it is not static as well because it is used by more than one function of the same swift file.

That variable does not need to be file-level.  Please scope it locally to `getRumSessionId` and think through the logic.   It does not need to be assigned in `createSessionIdChangeSpan`.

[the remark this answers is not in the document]
shattimare: make  `isSessionIdChanged` local variable and changed code  accordingly.

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -54,14 +56,25 @@ func getRumSessionId() -> String {
     }
     if Date() > sessionIdExpiration {
         sessionIdExpiration = Date().addingTimeInterval(TimeInterval(MAX_SESSION_AGE_SECONDS))
+        oldRumSessionId = rumSessionId
         rumSessionId = generateNewSessionId()
+        isSessionIdChanged = true
```
seemk, line 61: Can we just call `createSessionIdChangeSpan` here and get rid of the boolean?

shattimare: Let me explain what happens if we call` createSessionIdChangeSpan` in `getRumSessionId`.
In `getRumSessionId` when new session id is generated , we use `sessionIdLock.lock()`, if we call `createSessionIdChangeSpan` from `getRumSessionId` then agin it call `getRumSessionId`  to assign session id to newly created span , but `sessionIdLock` is already lock so dead lock situation will occur because of that app get crashed , so that we used `isSessionIdChanged` flag and after unlock the sessionIdLock , we Called `createSessionIdChangeSpan`

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -23,6 +23,7 @@ private var rumSessionId = generateNewSessionId()
 private var sessionIdExpiration = Date().addingTimeInterval(TimeInterval(MAX_SESSION_AGE_SECONDS))
 private let sessionIdLock = NSLock()
 private var sessionIdCallbacks: [(() -> Void)] = []
+private var oldRumSessionId = ""
```
johnbley, line 26: This can *also* be a local variable inside `getRumSessionId`.

johnbley (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
shattimare: keep `oldRumSessionId` in side `getRumSessionId`.

PR #58, opened 2022-02-17
Feature/apmi2306-Capture-screen-render-time

Piyush23786:
cadisplaylink for frozen frame


SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+        let frameDuration = lastFrameTimestamp - previousFrameTimestamp
+          // print(String(format: "%.02f", frameDuration))
+        if frameDuration > slowFrameThreshold && frameDuration <= frozenFrameThreshold {
+                reportSlowframe(e: frameDuration)
+        }
+
+        if frameDuration > frozenFrameThreshold {
+            reportSlowframe(e: frameDuration)
```
seemk, line 76: Frozen frames and slow frames should have different span names: `slowRenders`, `frozenRenders`.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,95 @@
...
+        let lastFrameTimestamp: CFTimeInterval = timestamp()
+            if previousFrameTimestamp == previousFrameInitalValue {
+                previousFrameTimestamp = lastFrameTimestamp
+                return
+            }
+
+        let frameDuration = lastFrameTimestamp - previousFrameTimestamp
+          // print(String(format: "%.02f", frameDuration))
+        if frameDuration > slowFrameThreshold && frameDuration <= frozenFrameThreshold {
+                reportSlowframe(e: frameDuration)
```
seemk, line 72: This is reported for every slow frame? The frames should be aggregated into 1s buckets

seemk (changes requested):

SplunkRumWorkspace/SmokeTest/SmokeTest/AppDelegate.swift
```diff
@@ -25,7 +25,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         // Override point for customization after application launch.
-        SplunkRum.initialize(beaconUrl: "http://127.0.0.1:8989/", rumAuth: "FAKE_RUM_AUTH", options: SplunkRumOptions(allowInsecureBeacon: true, debug: true, globalAttributes: [:], environment: nil, ignoreURLs: nil))
+        SplunkRum.initialize(beaconUrl: "https://rum-ingest.us0.signalfx.com/v1/rum", rumAuth: "nF2sRwMTyB-is8WpcGQ72w", options: SplunkRumOptions(allowInsecureBeacon: true, debug: true, globalAttributes: [:], environment: nil, ignoreURLs: nil))
```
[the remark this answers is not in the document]
Piyush23786: ok

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,95 @@
...
+        if frameDuration > frozenFrameThreshold {
+            reportSlowframe(e: frameDuration)
+        }
+ 
+     }
+    
+    func reportSlowframe(e: CFTimeInterval) {
+        let tracer = buildTracer()
+        let now = Date()
+        let typeName = "slow.frame"
```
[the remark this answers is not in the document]
Piyush23786: Done

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,95 @@
...
+            isRunning = false
+               // If we can't get the frame rate we assume it is 60.
+            var maximumFramesPerSecond = 60.0
+
+        if #available(iOS 10.3, *) {
+            maximumFramesPerSecond = Double(UIScreen.main.maximumFramesPerSecond)
+        }
+        slowFrameThreshold = 1 / (maximumFramesPerSecond - 1)
+
+        print(slowFrameThreshold)
```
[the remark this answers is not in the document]
Piyush23786: Done

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: added different spans for Frozen frames.

seemk:
Also let's please make the thresholds configurable

Piyush23786:
> Also let's please make the thresholds configurable

thresholds value is already configured.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: yes it's for every slow frame.

seemk:
> > Also let's please make the thresholds configurable
> 
> thresholds value is variable defined in the code. do we want to be user configurable?

yes, via options

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
seemk: We should limit the amount of slow render reports to 1 per second and also include the count of slow renders during this 1 second interval.

Piyush23786:
> > > Also let's please make the thresholds configurable
> > 
> > 
> > thresholds value is variable defined in the code. do we want to be user configurable?
> 
> yes, via options

A new option variable, framesPerSecond, has been added

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: interval 1 => fps: double 60 
interval 2 => fps: double 30 
interval 3 => fps: double 20 
**the interval goes from 1 to 60** 

like this way?

Piyush23786: As discussed only slow frame spans we created. CADisplayLink will Starts on viewDidAppear and stop on viewDidDisappear.

SplunkRumWorkspace/SmokeTest/SmokeTest/Base.lproj/Main.storyboard
```diff
@@ -31,7 +32,7 @@
                 </viewController>
                 <placeholder placeholderIdentifier="IBFirstResponder" id="dkx-z0-nzr" sceneMemberID="firstResponder"/>
             </objects>
-            <point key="canvasLocation" x="-196" y="41"/>
+            <point key="canvasLocation" x="-468.11594202898556" y="54.910714285714285"/>
```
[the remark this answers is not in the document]
Piyush23786: no this is not related to PR.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -37,14 +37,15 @@ let SplunkRumVersionString = "0.5.2"
...
-    @objc public init(allowInsecureBeacon: Bool = false, debug: Bool = false, globalAttributes: [String: Any] = [:], environment: String? = nil, ignoreURLs: NSRegularExpression? = nil, screenNameSpans: Bool = true) {
+    @objc public init(allowInsecureBeacon: Bool = false, debug: Bool = false, globalAttributes: [String: Any] = [:], environment: String? = nil, ignoreURLs: NSRegularExpression? = nil, screenNameSpans: Bool = true, framesPerSecond: Int = 0) {
         // rejectionFilter not specified to make it possible to call from objc
         self.allowInsecureBeacon = allowInsecureBeacon
         self.debug = debug
         self.globalAttributes = globalAttributes
         self.environment = environment
         self.ignoreURLs = ignoreURLs
         self.screenNameSpans = screenNameSpans
+        self.framesPerSecond = framesPerSecond
```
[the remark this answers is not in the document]
Piyush23786: we have configured CFTimeInterval as a threshold?

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,99 @@
...
+        displayLink?.invalidate()
+        displayLink = nil
+    }
+
+    @objc func displayLinkCallback(_ displayLink: CADisplayLink) {
+
+        duration = displayLink.targetTimestamp - displayLink.timestamp
+        if duration > slowFrameThreshold {
+            stopTracking()
+            reportSlowframe(slowFrame: duration)
```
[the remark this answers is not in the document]
Piyush23786: 15
15
Slow render detected: 16ms 15times
__________________
Span slowRenders:
TraceId: d7c0a7e3e0a3b13c5461c956f46d19be
SpanId: e17dfac233a867d0
Span kind: internal
TraceFlags: TraceFlags{sampled=true}
TraceState: TraceState(entries: [])
ParentSpanId: b8aa8d96d706666b
Start: 1646742753823909888
Duration: 0 nanoseconds
Attributes: ["app": TestStoryboardApp, "count": 14, "splunk.rumVersion": 0.5.2, "os.version": 15.2, "splunk.rumSessionId": 0518f657e2f8805cdbead5f74618873d, "device.model.name": Simulator (iPod touch (7th generation)), "thread.name": main, "screen.name": ViewController, "os.name": iOS, "app.version": 1.0, "component": ui]
------------------

like this way right?

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,94 @@
...
+    override init() {
+            super.init()
+            isRunning = false
+    }
+
+    func startTracking() {
+        isRunning = true
+        stopTracking() /// make sure to stop a previous running display link
+        let displayLink = CADisplayLink(target: self, selector: #selector(displayLinkCallback))
+        displayLink.preferredFramesPerSecond = SplunkRum.configuredOptions?.framesPerSecond ?? 0  /// (optional) if you do not define , then device uses maxFramesPS
```
seemk, line 43: I think we can always have it run at max frames per second (the default behaviour)

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,94 @@
...
+        self.displayLink = displayLink
+    }
+
+    func stopTracking() {
+        isRunning = false
+        displayLink?.invalidate()
+        displayLink = nil
+    }
+
+    @objc func displayLinkCallback(_ displayLink: CADisplayLink) {
```
seemk, line 54: How does this function when the application is in background? Is it guaranteed to not have huge jumps?

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,94 @@
...
+
+         if self.startedTime == 0.0 {
+            self.startedTime = CFAbsoluteTimeGetCurrent()
+            return
+         }
+
+         let currentTime: CFTimeInterval = CACurrentMediaTime()
+         let elapsedTime = currentTime - startedTime
+
+         let count = 1 / (displayLink.targetTimestamp - displayLink.timestamp)
```
seemk, line 64: This gives the FPS corresponding to this 1 frame time. We need to count the amount of slow frames in a 1 second period/bucket.

seemk (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: That's an optional feature I added for testing.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: During background mode, this function stops the loop cadisplay link.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: We count frames in 1-second intervals only. 
let actualFramesPerSecond = 1 / (displaylink.targetTimestamp - displaylink.timestamp)

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
seemk: But why do we have it in the final code?

Piyush23786: removed optional

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @objc func displayLinkCallback(_ displayLink: CADisplayLink) {
+
+         if self.startedTime == 0.0 {
+            self.startedTime = CFAbsoluteTimeGetCurrent()
+            return
+         }
+         let currentTime: CFTimeInterval = CACurrentMediaTime()
+         let elapsedTime = currentTime - startedTime
```
seemk, line 60: Is start time ever reset? It seems `elapsedTime` is always icnreasing, no?

seemk (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    @objc func displayLinkCallback(_ displayLink: CADisplayLink) {
+
+         let currentTime: CFTimeInterval = CACurrentMediaTime()
+         if isFirstIteration {
+            previousTimestamp = displayLink.targetTimestamp
+            isFirstIteration = false
+         }
+         let duration = displayLink.timestamp - previousTimestamp
+         previousTimestamp = displayLink.timestamp
```
seemk, line 65: I think we have to account for the application going background here. If the displaylink callback is not being invoked in background, we can have a case where the render time is huge.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,106 @@
...
+limitations under the License.
+*/
+
+import Foundation
+import QuartzCore
+import UIKit
+
+class ScreenFrames: NSObject {
+
+    private var isRunning = false
```
seemk, line 24: This is unused

Piyush23786: This is not in use, so I will remove it

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: Application state is used to do this when an application comes from the background to the foreground.

seemk: Can you elaborate?

Piyush23786: We modified the code based on your suggestion to manage the app going into the background. You can refer the code here: https://github.com/signalfx/splunk-otel-ios/pull/58/files#diff-ce6f5d7a4bad04dd0be70bbfda633a95a580402a8776c0fca2c0e859977248aaR53-R56

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    func stopTracking() {
+
+        displayLink?.invalidate()
+        displayLink = nil
+    }
+
+    @objc func displayLinkCallback(_ displayLink: CADisplayLink) {
+
+         let state = UIApplication.shared.applicationState
```
seemk, line 53: Instead of checking for the state here and creating a new CADisplayLink on every VC appearance, would it work to use `applicationDidEnterBackground` and `applicationDidBecomeActive` callbacks instead?

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,104 @@
...
+         if state == .background || state == .inactive {
+            isFirstIteration = true
+         }
+         let currentTime: CFTimeInterval = CACurrentMediaTime()
+         if isFirstIteration {
```
seemk, line 58: Minor, but on the first iteration when have nothing to compare against should we just set the previous timestamp to the current timestamp and bail out here instead? While setting previousTimestamp to `targetTimestamp` doesn't affect the logic, it will be easier to understand what's going on. I had a bit of difficulty understanding why `targetTimestamp`, which shows when the next frame displays, is used here.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,104 @@
...
+             if slowCount > 0 {
+                 reportSlowframe(slowFrameCount: slowCount, name: "slowRenders")
+             }
+
+             if frozenCount > 0 {
+                 reportSlowframe(slowFrameCount: frozenCount, name: "frozenRenders")
+             } else {
+                 if duration > frozenFrameThreshold {
+                     reportSlowframe(slowFrameCount: 1, name: "frozenRenders")
+                 }
```
seemk, line 85: The outer else branch is taken at the beginning of a new full second. Our goal was to bucket the counts to 1 second intervals. This line seems to report a frozen frame at the beginning of a new second. Now if there will be any additional frozen frames during the next second, there will be 2 spans generated instead of 1.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,104 @@
...
+             }
+             slowCount = 0
+             frozenCount = 0
```
seemk, line 88: Related to the previous comment, but the first frame of this full second can be slow/frozen as well so we might want to to count it.

seemk:
Also it might make sense to configure the thresholds via user options

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: @seemk, there will only be one span generated per second with this logic. You are correct that the previous span will be sent at the beginning of the new second. So if there are additional frozen frames generated during the next second that span will be sent in the subsequent iteration. Let me know if I am missing something here. We can take a short call to discuss this and conclude.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
Piyush23786: @seemk, sure let me try this and update the PR.

README.md
```diff
@@ -58,6 +58,8 @@ or
 | spanRejectionFilter | ((SpanData) -> SpanData?)? | Closure to modify or reject/ignore spans.  See example below.  | `nil` |
 | showVCInstrumentation | Bool | Enable span creation for ViewController Show events (not applicable to all UI frameworks/apps) | true |
 | screenNameSpans | Bool | Enable span creation for changes to the screen name | true |
+| slowFrameThreshold | CFTimeInterval | The SlowFrame Threshold is an optional configuration | 0.0169 |
+| frozenFrameThreshold | CFTimeInterval | The frozenFrame Threshold is an optional configuration | 0.7 |
```
theletterf, line 62: These descriptions are not useful, @Piyush23786. What do these threshold control?

+CC @seemk

theletterf (changes requested):

README.md
Piyush23786: The SlowFrame Threshold is an optional configuration that marks all the frames that took more than the specified time as slow frames. The user needs to provide this value in milliseconds.

theletterf: That's great, @Piyush23786 . Could you please edit the descriptions in the table to match that?

Piyush23786: Done

theletterf: I still see the file as before.

Piyush23786: https://github.com/signalfx/splunk-otel-ios/pull/58/files

theletterf (approved):

README.md
```diff
@@ -58,6 +58,8 @@ or
 | spanRejectionFilter | ((SpanData) -> SpanData?)? | Closure to modify or reject/ignore spans.  See example below.  | `nil` |
 | showVCInstrumentation | Bool | Enable span creation for ViewController Show events (not applicable to all UI frameworks/apps) | true |
 | screenNameSpans | Bool | Enable span creation for changes to the screen name | true |
+| slowFrameThreshold | CFTimeInterval | The SlowFrame Threshold is an optional configuration that marks all the frames that took more than the specified time as slow frames. User needs to provide this value in milliseconds. | 0.0169 |
+| frozenFrameThreshold | CFTimeInterval | The frozenFrame Threshold is an optional configuration that marks all the frames that took more than the specified time as frozen frames. User needs to provide this value in milliseconds. | 0.7 |
```
seemk, line 62: The values should be 16.7 and 700 ms

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,113 @@
...
+             if duration > frozenFrameThreshold {
+                 frozenCount += 1
+              } else if duration > slowFrameThreshold {
+                 slowCount += 1
+              }
```
seemk, line 97: This if can be hoisted out of the outer if/else, it's used in both branches

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -99,6 +103,16 @@ let SplunkRumVersionString = "0.5.2"
...
 
+    /**
+     The SlowFrame Threshold is an optional configuration
+     */
+    @objc public var slowFrameThreshold: CFTimeInterval = 0.0169
+
+    /**
+     The frozenFrame Threshold is an optional configuration
+     */
+    @objc public var frozenFrameThreshold: CFTimeInterval = 0.7
```
seemk, line 114: Please have the config options as milliseconds

johnbley (changes requested):
Where is the automated test that shows that a slow frame is reported correctly?  E.g., a button in the smoke test (before it goes down the browser integration rabbit hole) that produces such a span.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
Piyush23786: Done with threshold config options now in milliseconds.

README.md
Piyush23786: > The values should be 16.7 and 700 ms

Added.

seemk (approved):

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,109 @@
...
+
+         if currentIteration != iteration {
+
+             if slowCount > 0 {
+                 reportSlowframe(slowFrameCount: slowCount, name: "slowRenders")
+             }
+
+             if frozenCount > 0 {
+                 reportSlowframe(slowFrameCount: frozenCount, name: "frozenRenders")
+             } else {
```
johnbley, line 86: This `else` block is dead and should not be here - previous code already checks for `duration > frozenFrameThreshold`.

johnbley (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
```diff
@@ -0,0 +1,107 @@
...
+         }
+
+         if currentIteration != iteration {
+
+             if slowCount > 0 {
+                 reportSlowframe(slowFrameCount: slowCount, name: "slowRenders")
+             }
+
+             if frozenCount > 0 {
+                 reportSlowframe(slowFrameCount: frozenCount, name: "frozenRenders")
```
seemk, line 86: The reporting happens on the next second. What if the count was increased during the previous second while a different screen was active? The screen name will be reported wrongly?

seemk (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenFrames.swift
seemk: Consider this scenario:

```
time=0.5s | User is on screen A, there is a slow frame, slowCount = 1
time=0.6s | User moves to screen B
time=1.0s | currentItartion != iteration part is now true -> reportSlowFrame uses `getScreenName` which returns screen B
```

Piyush23786: Solved issue with screen name reported incorrectly.

seemk:
Closing in favor of https://github.com/signalfx/splunk-otel-ios/pull/90

Feel free to review / give ideas

PR #61, opened 2022-03-01
feature/APMI-2820-Screen name not proper

shattimare:
Implemented code to show screen name on which crash occur.


seemk (commented):
Can we get rid of the unnecessary changes? Will keep a cleaner commit history then. If necessary the packages can be updated in a separate PR.

shattimare:
> Can we get rid of the unnecessary changes? Will keep a cleaner commit history then. If necessary the packages can be updated in a separate PR.

I raised separate PR.

PR #66, opened 2022-03-07
feature/APMI-2820

shattimare:
Implemented code.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
@@ -47,13 +47,13 @@ func internal_setScreenName(_ newName: String, _ manual: Bool) {
     if manual || !screenNameManuallySet {
         screenName = newName
     }
+    userDefaults.set(screenName, forKey: SCREEN_NAME_KEY)
     lock.unlock()
 
     // Don't emit the span under the lock
     if oldName != nil && oldName! != "unknown" && (SplunkRum.configuredOptions?.screenNameSpans ?? true) {
         emitScreenNameChangedSpan(oldName!, newName)
     }
-    userDefaults.set(screenName, forKey: SCREEN_NAME_KEY)
 }
 
 func isScreenNameManuallySet() -> Bool {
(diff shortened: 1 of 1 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
```diff
@@ -52,6 +53,7 @@ func internal_setScreenName(_ newName: String, _ manual: Bool) {
     if oldName != nil && oldName! != "unknown" && (SplunkRum.configuredOptions?.screenNameSpans ?? true) {
         emitScreenNameChangedSpan(oldName!, newName)
     }
+    userDefaults.set(screenName, forKey: SCREEN_NAME_KEY)
```
seemk, line 56: Should this perhaps be `newName` here? `screenName` is being accessed outside of a lock

seemk: or at least protect the access

shattimare: We just save ScreenName in userdefaults nothing else. if we want to access screen name we have to use getScreenName().

shattimare: I added userDefaults.set(screenName, forKey: SCREEN_NAME_KEY) with in  Lock block.

johnbley:
As I read this code, it seems to enact that "the first screen name before anything explicitly happens in the UI will have whatever was the last screen name during the previous run of the app".  This "fixes" the "I don't know what the screen name was during a crash" but at the expense of polluting startup spans with data of dubious quality.  A better way to do this is to update this data on the `crashReporter.customData` as we do for session IDs, which will involve (a) emitting and decoding a more complex structure than just the (string) session ID that is used for `customData` now, (b) an interface for the crash reporting module to use to be notified of screen name changes (similar to the session ID change api), and (c) work across the two repos.  Note that "tell the crash reporter about a screen name change" can't be hard coded in this (splunk-otel-ios) repo due to the strong separation between them.

shattimare:
We solve the issue using  way suggested by john, For that we had to make change in splunkrum and crashreporting library as well. We also raised PR of splunk-otel-ios-crashreporting.

PR #68, opened 2022-03-16
feature/APMI-2820 iOS crash does not retain old screen name

shattimare:
Changes in screenname and splunkrum

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
@@ -49,7 +49,6 @@ func internal_setScreenName(_ newName: String, _ manual: Bool) {
     }
     callbacks = screenNameCallbacks
     lock.unlock()
-    
     for callback in callbacks {
         callback()
     }
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
@@ -17,12 +17,13 @@ limitations under the License.
 
 import Foundation
 
+let userDefaults = UserDefaults.standard
 fileprivate var screenName: String = "unknown"
 fileprivate var screenNameManuallySet = false
 // Yes, I assume there are swift libraries to do this sort of thing, but nothing
 // I could find in the stdlib
 fileprivate var lock = NSLock()
-private var screenNameCallbacks: [(() -> Void)] = []
+private var screenNameCallbacks: [((String) -> Void)] = []
 
 func emitScreenNameChangedSpan(_ oldName: String, _ newName: String) {
     let now = Date()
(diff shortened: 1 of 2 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
```diff
@@ -22,6 +22,7 @@ fileprivate var screenNameManuallySet = false
 // Yes, I assume there are swift libraries to do this sort of thing, but nothing
 // I could find in the stdlib
 fileprivate var lock = NSLock()
+private var screenNameCallbacks: [(() -> Void)] = []
```
seemk, line 25: Could we just pass the screen name as a parameter to the callbacks? Could get rid of `getCurrentScreenName` then

shattimare: No, we are not passing just screen name as a parameter to the callbacks, `updateCrashReportScreenName` of crashreporting library will pass as parameter of it. when ever screen name will changed in splunkRum libaray, this method  will be executed and will updated screen name  in crashreporting libaray's `custom data`. with in `updateCrashReportScreenName` method it will called getCurrentScreenName  to get current screen name.  `getCurrentScreenName` is used to communicate with crashreporting library. I had railsed PR of crashreporting  library as well.

johnbley: I agree with Siim - passing the new value as a parameter of the callback would allow us to drop `getCurrentScreenName`.

shattimare: if `sessionbasedsamplingratio` is out of range (0.0 , 1.0) we consider it as 100% and sampled all session. as we use 1.0 as default value. if you want any change in it then let us know.

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
```diff
@@ -17,11 +17,13 @@ limitations under the License.
 
 import Foundation
 
+let userDefaults = UserDefaults.standard
```
seemk, line 20: This is unused

SplunkRumWorkspace/SplunkRum/SplunkRum/ScreenName.swift
```diff
@@ -17,11 +17,13 @@ limitations under the License.
...
 fileprivate var screenName: String = "unknown"
 fileprivate var screenNameManuallySet = false
 // Yes, I assume there are swift libraries to do this sort of thing, but nothing
 // I could find in the stdlib
 fileprivate var lock = NSLock()
+private var screenNameCallbacks: [((String) -> Void)] = []
```
seemk, line 26: fileprivate for consistency?

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -215,14 +215,18 @@ var splunkRumInitializeCalledTime = Date()
...
      Adds a callback whenever the sessionId changes.
      */
     public class func addSessionIdChangeCallback(_ callback: @escaping (() -> Void)) {
         addSessionIdCallback(callback)
     }
-
+    /**
+     Adds a callback whenever the screen name changes.
+     */
+    public class func addScreenNameChangeCallback(_ callback: @escaping ((String) -> Void)) {
```
seemk, line 227: Where is this called from in this PR?

seemk: My bad, it was in the crash reporter

seemk (approved):

PR #69, opened 2022-03-21
Feature/apmi 2771   improved app start up measurement

shattimare:
Implemented code. 
We should initialise Splunk RUM in applicationWillResignActive method as well


seemk (commented):
A small overview or an explanation would be nice here.

Would it also be testable?

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -178,25 +178,36 @@ var splunkRumInitializeCalledTime = Date()
...
-        }
-        sendAppStartSpan()
-        let srInit = buildTracer()
-            .spanBuilder(spanName: "SplunkRum.initialize")
-            .setStartTime(time: splunkRumInitializeCalledTime)
-            .startSpan()
-        srInit.setAttribute(key: "component", value: "appstart")
-        if options != nil {
-            srInit.setAttribute(key: "config_settings", value: options!.toAttributeValue())
+        DispatchQueue.global().async {
```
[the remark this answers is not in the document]
shattimare: > Can you please give a small overview why is this done in a global queue?

**Problem:** Lets say, User has started the app , the initialisation method started but was not completed ,before that user put the app in background mode, in that case, might be the initialisation of Splunk SDK was not fully initialised 

**Solution:**  Implemented the background thread to run initialisation code in above situation.

SplunkRumWorkspace/SplunkRum/SplunkRum.xcodeproj/project.pbxproj
```diff
@@ -511,7 +511,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
```
seemk, line 514: Is there a reason for changing the deployment target from 11 to 12?

seemk (commented):
Need to remove the unrelated changes (dependency updates, min version). Otherwise I think it should be ok to do it like thiiis

SplunkRumWorkspace/SplunkRum/SplunkRum.xcodeproj/project.pbxproj
shattimare: @seemk  for testing purpose i did that.

PR #70, opened 2022-03-23
Feature/apmi 2313 implement disk caching

shattimare:
Implemented code.

There are two points to clarify.
1. When we generate span from cache data, we are not able to use cached value of span id , traceid, traceFlags, traceState, parentSpanId to export it on RUM playground. Reason of it is that setter method of that property is private, so we are not able to set it. It will randomly generate spanID, traceID etc.
2. After network will available , cached spans will not be exported automatically on playground instead cached span will be exported to playground with span which will generate by user action.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/DiskCache/CoreDataManager.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/DiskCache/CoreDataManager.swift
@@ -23,7 +23,7 @@ import OpenTelemetrySdk
 import SwiftUI
 
 let FLUSH_OUT_TIME_SECONDS = 60   // 4 * 60 * 60  // 4 hours
-
+let FLUSH_OUT_MAX_SIZE = 2e+7  // 20 MB
 
 let Entity_name = "Pending"
 let TimeStampColumn = "created_at"
@@ -40,10 +40,14 @@ public class CoreDataManager {
             let messageKitBundle = Bundle(identifier: self.identifier)
             let modelURL = messageKitBundle!.url(forResource: self.model, withExtension: "momd")!
             let managedObjectModel =  NSManagedObjectModel(contentsOf: modelURL)
-            
+        
+            let storedescription = NSPersistentStoreDescription(url: NSPersistentContainer.defaultDirectoryURL().appendingPathComponent("Rum.sqlite"))
+            storedescription.setOption(true as NSNumber, forKey: NSSQLiteManualVacuumOption)
+       
             let container = NSPersistentContainer(name: self.model, managedObjectModel: managedObjectModel!)
+            container.persistentStoreDescriptions = [storedescription]
             container.loadPersistentStores { (storeDescription, error) in
-                
+                 
                 if let err = error{
                     fatalError("Loading of store failed:\(err)")
                 }
@@ -52,12 +56,9 @@ public class CoreDataManager {
             return container
         }()
     
-    
-    
-    //MARK:- Insert span in to DB
-
+//MARK:- Insert span in to DB
 public func insertSpanIntoDB(_ spans: [SpanData]) {
-   // getStoreInformation()
+    getStoreInformation()
    
     let spanArr : Array = [1,2,3,4]
    // let spanArr : Array = [spans]
@@ -86,9 +87,11 @@ public func fetchSpanFromDB() -> [SpanData] {
         let result1 = try managedObjectContext.fetch(fetchRequest) as! [Pending]
         for data in result1 {
            // guard let spanDataArray = (try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(data.span!)) as? [SpanData] else { return [] }
-            guard let spanDataArray = (try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(data.span!)) else { return [] }
+            guard (try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(data.span!)) != nil else { return [] }
            
-            print("out put is/////////// \(spanDataArray)")
+           /* guard let spanDataArray = (try? NSKeyedUnarchiver.unarchiveTopLevelObjectWithData(data.span!)) else { return [] }
+           
+            print("out put is/////////// \(spanDataArray)")*/
            
           
         }
@@ -117,6 +120,7 @@ public func fetchSpanFromDB() -> [SpanData] {
         }*/
         return result
     }
+    //MARK: - Flush out logic -
     /** flush out db after 4 h time out */
     public func flushOutSpanAfterTimePeriod() {
         print("Core Data Store at ......\(NSPersistentContainer.defaultDirectoryURL().absoluteString)")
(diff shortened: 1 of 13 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -180,7 +180,7 @@ var splunkRumInitializeCalledTime = Date()
         let limiting = LimitingExporter(proxy: retry, spanFilter: options?.spanFilter ?? nil)
         OpenTelemetrySDK.instance.tracerProvider.addSpanProcessor(BatchSpanProcessor(spanExporter: limiting))
         if options?.debug ?? false {
-            OpenTelemetrySDK.instance.tracerProvider.addSpanProcessor(SimpleSpanProcessor(spanExporter: StdoutExporter(isDebug: true)))
+            OpenTelemetrySDK.instance.tracerProvider.addSpanProcessor(SimpleSpanProcessor(spanExporter: StdoutExporter(isDebug: false)))
```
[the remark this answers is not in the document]
shattimare: isDebug should be true

SplunkRumWorkspace/SplunkRum/SplunkRum/RetryExporter.swift
```diff
@@ -29,45 +29,41 @@ class RetryExporter: SpanExporter {
...
-        pending.append(contentsOf: spans)
-        if pending.count > MAX_PENDING_SPANS {
-            pending.removeFirst(pending.count - MAX_PENDING_SPANS)
-        }
-    }
-
-    func export(spans: [SpanData]) -> SpanExporterResultCode {
-        var result = attemptPendingExport()
+  func export(spans: [SpanData]) -> SpanExporterResultCode {
+        var result = attemptDBExport()
```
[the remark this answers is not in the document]
shattimare: When network is not available , we store all the spans generate during that time into DB. All this stored span will export to rum portal once network is available and any action or new span will be generated.

attemptDBExport function fetch span information from db and create new span using that information and export that spans. This function return success. In next line `result = proxy.export(spans: spans)` will send current span on playground.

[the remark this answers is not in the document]
seemk: I might be missing something, but I can't see where the spans loaded from disk are exported. It seems `attemptDBExport` fetches the spans, but does not export them anywhere?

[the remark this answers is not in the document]
shattimare: All db related stuffs are available in `CoreDataManager` class.In it there is a `fetchSpanValues` function, it  fetch span information from db and generate new span using  `createNewSpan` function, In `createNewSpan` function we use span builder to create new span using information from db and at the end we use `span.end(time: spanInfo.value(forKey: "duration") as! Date)` to end the span , as we end the span , span is added into the exporter queue(this logic is already developed in SDK) and network is available than it send to RUM playground using (Exporter’s chain) `ZipkinTraceExporter’s` export method.  If the span send successfully then delete that spans from db else not.

SplunkRumWorkspace/SplunkRum/SplunkRumTests/UtilsTests.swift
```diff
@@ -196,12 +196,12 @@ class UtilsTests: XCTestCase {
...
         _ = re.export(spans: fiftySpans)
         _ = re.export(spans: fiftySpans)
-        XCTAssertEqual(0, localSpans.count)
+        XCTAssertEqual(150, localSpans.count)
 
         // now reconnected; next export of 50 should send total of 150 (last 100 retried)
         te.exportSucceeds = true
         _ = re.export(spans: fiftySpans)
-        XCTAssertEqual(150, localSpans.count)
+        XCTAssertEqual(250, localSpans.count)
```
seemk, line 204: Perhaps we need additional tests besides this?

SplunkRumWorkspace/SplunkRum/SplunkRum/DiskCache/CoreDataManager.swift
```diff
@@ -0,0 +1,279 @@
...
+                span.setAttribute(key: dict.key, value: value as! Int)
+            } else if dict.value is Double {
+                span.setAttribute(key: dict.key, value: value as! Double)
+            } else if dict.value is Bool {
+                span.setAttribute(key: dict.key, value: value as! Bool)
+            } else if dict.value is [String: Any] {
+                span.setAttribute(key: dict.key, value: value as? AttributeValue)
+            }
+        }
+        span.end(time: spanInfo.value(forKey: "duration") as! Date)
```
seemk, line 130: Is the `SpanContext` lost here?

shattimare: @seemk, Here we are creating custom span using information stored in database, when we create custom span I think we are not using SpanContext.

SplunkRumWorkspace/SplunkRum/SplunkRumTests/UtilsTests.swift
shattimare: @seemk added additional tests.

seemk: Can we verify that serialized attributes are all loaded correctly? I.e. when constructing spans from disk

shattimare: Add test case to check that attributes are loaded properly.

SplunkRumWorkspace/SplunkRum/SplunkRum/DiskCache/CoreDataManager.swift
shattimare: @seemk We can create SpanContext using traceId, spanID,t raceFlags, traceState. As mention in my very first  comment traceId, spanId, traceFlags are private property, So we can’t able to access it and create span context using that information stored on disk.

seemk (changes requested):
There are a few problems with the current approach:
1. Spans are loaded from disk when a span is attempted to be exported. It would be better to do it continuously in the background.
2. There seems to be no limit to the amount of spans fetched - the spans loaded from disk will go through the same exporting loop, meaning the throttle will drop spans loaded from disk.

shattimare:
> There are a few problems with the current approach:
> 
> 1. Spans are loaded from disk when a span is attempted to be exported. It would be better to do it continuously in the background.
> 2. There seems to be no limit to the amount of spans fetched - the spans loaded from disk will go through the same exporting loop, meaning the throttle will drop spans loaded from disk.

1. Implemented code for exporting cached span continuously when network is available. Did the code for it in `Networktype.swift` file.
2. Set limit of span fetching from disk, currently set it to 2 for testing purpose , we can change it later on.

shattimare:
As per discussed with @seemk closing this PR.
he created separate branch and added this logic and merge to main branch.

PR #73, opened 2022-03-30
Feature/apmi 2630 session based sampling

shattimare:
Implemented code.

Overview:

We have to pass `sessionBaseSamplingRatio` value between 0.0 to 1.0 when initialise splunkrum from app delegate.
It’s default value is 1.0. In `initialize` method  of `SplunkRum` class , We checked that if value is with in above range than pass this value to `SessionBasedSampling` class. Call `sessionshouldSample` method of it. In this method we developed algorithm to check that current session will be sampled or not, and set value of result variable accordingly.
If result is true means session will be sampled and false means session will not be sampled.

At the end set the value of active sampler. We use timer to call `sessionshouldSample` method  whenever new session will be generated.

In order to handle 15 min inactivity session time out ,call `sessionshouldSample` from `invalidateSession` method of AppLifeCycle class. I had created branch from APMI-2307 (session time out after 15 min user in activity) than raised PR.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
@@ -19,6 +19,9 @@ import Foundation
 import UIKit
 import OpenTelemetrySdk
 
+let INACTIVITY_SESSION_TIMEOUT_SECONDS = 15 * 60
+private var sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+
 func initializeAppLifecycleInstrumentation() {
     /*
      Observed event sequences:
@@ -57,6 +60,7 @@ func initializeAppLifecycleInstrumentation() {
 }
 var activeSpan: SpanHolder?
 func lifecycleEvent(_ event: String) {
+    invalidateSession(event)
     // these two start spans
     if event == "UIApplicationWillResignActiveNotification" ||
             event == "UIApplicationWillEnterForegroundNotification" {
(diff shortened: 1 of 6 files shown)


shattimare:
if `sessionbasedsamplingratio` is out of range (0.0 , 1.0) we consider it as 100% and sampled all session. as we use 1.0 as default value. if you want any change in it then let us know.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
```diff
@@ -19,6 +19,10 @@ import Foundation
 import UIKit
 import OpenTelemetrySdk
 
+let INACTIVITY_SESSION_TIMEOUT_SECONDS = 15 * 60
+private var sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+var Is_New_SessionID_For_In_Activity = false
```
seemk, line 24: Would be good to have consistent style for variables

SplunkRumWorkspace/SplunkRum/SplunkRum/SessionBasedSampling.swift
```diff
@@ -0,0 +1,78 @@
...
+            } else {
+                result = false
+            }
+        }
+
+        var psampler: Sampler
+        if result {
+                psampler = OpenTelemetrySdk.Samplers.parentBased(root: Samplers.alwaysOn, remoteParentSampled: Samplers.alwaysOn, remoteParentNotSampled: Samplers.alwaysOn, localParentSampled: Samplers.alwaysOn, localParentNotSampled: Samplers.alwaysOn)
+        } else {
+              psampler = OpenTelemetrySdk.Samplers.parentBased(root: Samplers.alwaysOff, remoteParentSampled: Samplers.alwaysOff, remoteParentNotSampled: Samplers.alwaysOff, localParentSampled: Samplers.alwaysOff, localParentNotSampled: Samplers.alwaysOff)
```
seemk, line 54: Wouldn't it be easier to use the `Sampler` protocol and set session based sampler as the active sampler in tracer provider?

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
shattimare: We create this PR from branch of APMI-2307 `https://github.com/signalfx/splunk-otel-ios/pull/51` , so this change is already implemented in it , when it will be approved and merged this problem might be solved.

SplunkRumWorkspace/SplunkRum/SplunkRum/SessionBasedSampling.swift
shattimare: Yes, I had tried to use `alwaysOn` and `alwaysOff` sampler. But once active sampler’s value be `alwaysOff`, no session was sampled after that even though I changed value of active sampler to `alwaysOn`. As a solution , I use `parent based `sampler protocol in place of it.

PR #75, opened 2022-04-06
SWAT-4463

Piyush23786:
change the swift-tools-version:5.6


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.6
```
seemk, line 1: Can we try 5.5? It seems the runner does not have 5.6

PR #76, opened 2022-04-06
Swat-4463 iOS library cannot be imported successfully in xCode 13.3 via SPM

Piyush23786:
changed swit-tools-version 5.5


seemk:
Might it be necessary to use multiple Package.swift files if that is possible? Or is there another way to support multiple versions?
Don't think making everyone use the newer toolchain is the best path forward here

PR #80, opened 2022-04-27
Feature/Apmi-2899 method/function to check if the library has been initialized

Piyush23786:
method/function to check if the library has been initialized

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
@@ -147,14 +147,14 @@ var splunkRumInitializeCalledTime = Date()
                 - Parameter options: Non-required configuration toggles for various features.  See SplunkRumOptions struct for details.
      
      */
-    @objc public class func initialize(beaconUrl: String, rumAuth: String, options: SplunkRumOptions? = nil) {
+    @objc public class func initialize(beaconUrl: String, rumAuth: String, options: SplunkRumOptions? = nil) -> Bool {
         if !Thread.isMainThread {
             print("SplunkRum: Please call SplunkRum.initialize only on the main thread")
-            return
+            return false
         }
         if initialized || initializing {
             debug_log("SplunkRum already initializ{ed,ing}")
-            return
+            return false
         }
         splunkRumInitializeCalledTime = Date()
         initializing = true
@@ -173,7 +173,7 @@ var splunkRumInitializeCalledTime = Date()
         }
         if !beaconUrl.starts(with: "https:") && options?.allowInsecureBeacon != true {
             print("SplunkRum: beaconUrl must be https or options: allowInsecureBeacon must be true")
-            return
+            return false
         }
         if rumAuth.isEmpty {
             theBeaconUrl = beaconUrl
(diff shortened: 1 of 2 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -335,4 +335,11 @@ var splunkRumInitializeCalledTime = Date()
         integrateWebViewWithBrowserRum(view: view)
     }
 
+    /**
+       This check is to determine whether the splunkrum library has been initialized
+     */
+    public class func rumInitialize() -> Bool? {
```
seemk, line 341: Perhaps rename it to `isInitialized`, currently it looks as if it does the initialization

seemk (changes requested):
We also need `initialize` itself return a boolean: `true` on first init and `false` otherwise

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
Piyush23786: Done

seemk:
The goal here was to make `SplunkRum.initialize` return a boolean as well

seemk (approved):

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -335,4 +337,11 @@ var splunkRumInitializeCalledTime = Date()
         integrateWebViewWithBrowserRum(view: view)
     }
 
+    /**
+       This check is to determine whether the splunkrum library has been initialized
+     */
+    public class func isInitialized() -> Bool? {
```
johnbley, line 343: Please adjust the unit tests so that before the call to `initialize`, `isInitialized` returns false, and then that the call to `initialize` returns true, and that then `isInitialized` also returns true.

johnbley (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
Piyush23786: I will add unit tests.

SplunkRumWorkspace/SplunkRum/SplunkRumTests/TestUtils.swift
```diff
@@ -100,7 +100,10 @@ func initializeTestEnvironment() throws {
     options.environment = "env"
     options.ignoreURLs = try! NSRegularExpression(pattern: ".*ignore_this.*")
 
-    SplunkRum.initialize(beaconUrl: "http://127.0.0.1:8989/v1/traces", rumAuth: "FAKE", options: options)
+    let rumInitialize = SplunkRum.initialize(beaconUrl: "http://127.0.0.1:8989/v1/traces", rumAuth: "FAKE", options: options)
+    let isRUMInitialized = SplunkRum.isInitialized()
+    let initializeAgain = SplunkRum.initialize(beaconUrl: "http://127.0.0.1:8989/v1/traces", rumAuth: "FAKE", options: options)
+    let isStillInitialized = SplunkRum.isInitialized()
```
seemk, line 106: Could perhaps add the asserts here right away

Piyush23786: Added

johnbley (approved):

PR #82, opened 2022-06-01
feat: disk caching span export

seemk:
Summary

* New option to `SplunkRum.initialize`: `enableDiskCache`, will use an alternate span exporter pipeline.
* The new pipeline consists of:
  - `SpanToDiskExporter` - the OTel `SpanExporter` implementation, serializes spans to Zipkin JSON string and writes them to a local database. Internally uses `SpanDb`, an SQLite database. Configurable with a max database size (default = 64MB). The exporter periodically attempts to truncate the database after some amount of spans (2048) have been inserted. The truncation happens in the processor thread and only if the database exceeds its size limit. If the size limit is reached, 20% of the oldest spans are dropped and a vacuum is performed to take back disk space. In case of directory permission errors (e.g. when the application support directory can't be accessed, should not happen), the database is opened with a memory backing store, so uploading will still work, but is not persistent between runs.
  - `SpanFromDiskExporter` - a background thread which periodically polls the database for new spans and posts them to the configured endpoint **oldest first**. The spans are not deserialized, but the existing JSON spans are glued together. Span transfer is bandwidth-limited (15 KiB/s). The usual database poll interval is 5 seconds, but when the bandwidth limit is active, polling is done each second. After a successful export, the poll interval is removed to allow for burst uploads (while still respecting the bandwidth limit). When the spans have been exported, they are erased from the local database.
  
Notes

* Zipkin serialization code is copied and slightly modified version from the OpenTelemetry SDK. There might still be unnecessary fields being serialized, but this can be removed later on.
* OTel's Zipkin exporter does not handle server status codes, only when there is a transport error. I've also made the exporting to handle retriable and non-retriable status codes. Meaning HTTP timeouts, server errors and such won't drop the spans (which in the case of OTel Zipkin exporter did happen, e.g. HTTP Too Many Requests is interpreted as a successful response there).

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/SpanDb.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/SpanDb.swift
@@ -39,7 +39,7 @@ class SpanDb {
         var status = sqlite3_open(databasePath, &db_)
 
         if status != SQLITE_OK {
-            print("Failure opening \(databasePath): \(sqliteError(code: status))")
+            log("failure opening \(databasePath): \(sqliteError(code: status))")
             return
         }
 
@@ -48,28 +48,28 @@ class SpanDb {
         status = sqlite3_exec(db, "CREATE TABLE IF NOT EXISTS span (timestamp INTEGER NOT NULL, data TEXT NOT NULL)", nil, nil, nil)
 
         if status != SQLITE_OK {
-            print("Unable to create span table: \(sqliteError(code: status))")
+            log("unable to create span table: \(sqliteError(code: status))")
             return
         }
 
         status = sqlite3_exec(db, "CREATE INDEX IF NOT EXISTS span_timestamp ON span (timestamp)", nil, nil, nil)
 
         if status != SQLITE_OK {
-            print("Unable to create span_timestamp index: \(sqliteError(code: status))")
+            log("unable to create span_timestamp index: \(sqliteError(code: status))")
             return
         }
 
         status = sqlite3_prepare_v2(db, "INSERT INTO span (timestamp, data) VALUES (?, ?)", -1, &insertStmt_, nil)
 
         if status != SQLITE_OK {
-            print("Unable to create span insert statement: \(sqliteError(code: status))")
+            log("unable to create span insert statement: \(sqliteError(code: status))")
             return
         }
 
         status = sqlite3_prepare_v2(db, "SELECT rowid, data FROM span ORDER BY timestamp ASC LIMIT ?", -1, &fetchStmt_, nil)
 
         if status != SQLITE_OK {
-            print("Unable to create span fetch statement: \(sqliteError(code: status))")
+            log("unable to create span fetch statement: \(sqliteError(code: status))")
             return
         }
 
@@ -80,7 +80,7 @@ class SpanDb {
                                     nil)
 
         if status != SQLITE_OK {
-            print("Unable to create DB size fetch statement: \(sqliteError(code: status))")
+            log("unable to create DB size fetch statement: \(sqliteError(code: status))")
             return
         }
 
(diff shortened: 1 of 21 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/SpanDb.swift
```diff
@@ -0,0 +1,281 @@
...
+    private var sizeStmt_: OpaquePointer?
+    private var initialized: Bool = false
+
+    init(path: String? = nil) {
+        self.databasePath = path ?? SpanDb.makeDatabasePath() ?? ":memory:"
+
+        var status = sqlite3_open(databasePath, &db_)
+
+        if status != SQLITE_OK {
+            print("Failure opening \(databasePath): \(sqliteError(code: status))")
```
johnbley, line 42: I've tried to use the term `SplunkRum` in any log messages like this before (e.g., `print("SplunkRum: Please call SplunkRum.initialize only on the main thread")`.  Can you please audit these and make sure that anybody staring at some random logs can know this stuff came from us?

SplunkRumWorkspace/SplunkRum/SplunkRum/SpanDb.swift
```diff
@@ -0,0 +1,281 @@
...
+        defer {
+            lock.unlock()
+        }
+
+        // Delete the oldest 20% of spans
+        let query = """
+          DELETE FROM span WHERE rowid IN (
+            SELECT rowid FROM span ORDER BY timestamp ASC LIMIT (
+              SELECT COUNT(*) / 5 FROM span
+            )
```
johnbley, line 219: 😄 reads very cleanly!

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -181,11 +190,21 @@ var splunkRumInitializeCalledTime = Date()
...
         OpenTelemetrySDK.instance.tracerProvider.addSpanProcessor(GlobalAttributesProcessor())
-        let exportOptions = ZipkinTraceExporterOptions(endpoint: beaconUrl+"?auth="+rumAuth, serviceName: "myservice") // FIXME control zipkin better to not emit unneeded fields
-        let zipkin = ZipkinTraceExporter(options: exportOptions)
-        let retry = RetryExporter(proxy: zipkin)
-        let limiting = LimitingExporter(proxy: retry, spanFilter: options?.spanFilter ?? nil)
-        OpenTelemetrySDK.instance.tracerProvider.addSpanProcessor(BatchSpanProcessor(spanExporter: limiting))
+        let exportOptions = ZipkinTraceExporterOptions(endpoint: theBeaconUrl!, serviceName: "myservice") // FIXME control zipkin better to not emit unneeded fields
+
+        if options?.enableDiskCache ?? false {
+            let spanDb = SpanDb()
```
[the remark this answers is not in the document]
seemk: Good point, I added deletion of the database file when disk caching is disabled.

SplunkRumWorkspace/SplunkRum/SplunkRumDiskExportTests/DiskExportCommon.swift
```diff
@@ -0,0 +1,86 @@
+//
```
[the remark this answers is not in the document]
seemk: They were run in `Product->Test`, but were missing from `fullbuild.sh`. Took me a bit to find how to do it - had to re-toggle macOS at SplunkRum's deployment info section, creating a new target did copy macOS to it. 🤷

SplunkRumWorkspace/SplunkRum/SplunkRum/SpanDb.swift
seemk: 👍 Done: https://github.com/signalfx/splunk-otel-ios/pull/82/commits/22f4b97fbf3cda0dfe9990795ef24f4c1e9a140f

johnbley (approved):

johnbley:
Code looks good and well thought through; perhaps update the README with the new option unless we don't want to document this yet.

seemk:
> Code looks good and well thought through; perhaps update the README with the new option unless we don't want to document this yet.

Added a section for `enableDiskCache` to README

PR #90, opened 2022-08-07
feat: slow and frozen frame detection

seemk:
Rework of https://github.com/signalfx/splunk-otel-ios/pull/58

Changes:
* Pause `CADisplayLink` when on app will resign and resume it on app activation. Simplifies the code somewhat and avoids issues with counting the frame duration when the app is moved to background.
* Group slow and frozen frames per screen names. This is necessary so when 2 different screens have a slow frame in a second interval, they are both reported.
* Use screen name callbacks to store the screen name locally to avoid the locking in `getScreenName()`
* Use a 1 second timer to periodically dump the detected frames.


aurbiztondo-splunk (approved):

SplunkRumWorkspace/SplunkRum/SplunkRum/SlowFrameDetector.swift
```diff
@@ -0,0 +1,128 @@
...
+limitations under the License.
+*/
+
+import Foundation
+import QuartzCore
+import UIKit
+
+fileprivate var detector: SlowFrameDetector = SlowFrameDetector()
+
+class SlowFrameDetector: NSObject {
```
pablocornejo, line 24: Can this class be `final`?

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -119,6 +127,15 @@ public let DEFAULT_DISK_CACHE_MAX_SIZE_BYTES: Int64 = 25 * 1024 * 1024
      Enable NetworkInstrumentation span creation for https calls.
      */
     @objc public var networkInstrumentation: Bool = true
+     /**
+     Threshold, in milliseconds, from which to count a rendered frame as slow.
+     */
+    @objc public var slowFrameDetectionThresholdMs: Double = 16.7
```
pablocornejo, line 133: Would it make sense to make this dynamic based on the [max frames per second](https://developer.apple.com/documentation/uikit/uiscreen/2806814-maximumframespersecond) that the screen can render?

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -119,6 +127,15 @@ public let DEFAULT_DISK_CACHE_MAX_SIZE_BYTES: Int64 = 25 * 1024 * 1024
...
+
+    /**
+     Threshold, in milliseconds, from which to count a rendered frame as frozen.
+     */
+    @objc public var frozenFrameDetectionThresholdMs: Double = 700
```
pablocornejo, line 138: Can this be a `let`, or is the value changed at some point?

johnbley (approved):

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
johnbley: Options structures are designed to support the use case of creating one, setting various fields manually, and then passing it to the `initialize`.  We keep a copy internally so it won't change over the lifespan of the app.

johnbley:
Unfortunately I can't merge this right now as there were unsigned commits.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
johnbley: Hi Pablo - it's an interesting idea.  Let's get this PR back into shape, merge it, and then we can look at pursuing this as an additional tweak after.

theletterf:
Watching this one for docs updates. Ping me if a different PR is opened.

PR #91, opened 2022-08-11
APMI-2306 slow and frozen frame detection

Piyush23786:
Changes: with signed commits.

Pause CADisplayLink when on app will resign and resume it on app activation. Simplifies the code somewhat and avoids issues with counting the frame duration when the app is moved to background.
Group slow and frozen frames per screen names. This is necessary so when 2 different screens have a slow frame in a second interval, they are both reported.
Use screen name callbacks to store the screen name locally to avoid the locking in getScreenName()
Use a 1 second timer to periodically dump the detected frames.


johnbley:
You can definitely disable that line length warning in the `.swiftline.yml`.

Piyush23786:
> 

ok

Piyush23786:
all changes from #90 with signed commit.

johnbley (approved):

PR #94, opened 2022-09-06
APMI-3255:- iOS: options factory update

Piyush23786:
APMI-3255:-
1. Set Location method that sets global attributes (only 2 numbers - latitude and longitude)
2. Bundle identifier (CF Bundle) that is detected and added automatically.
3. Feature flag for slow/frozen renders
4. Contents attribute added.


SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumOptions.swift
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     Threshold in bytes from which spans will start to be dropped from the disk cache (oldest first).
+     Only applicable when disk caching is enabled.
+     */
+    @objc public var spanDiskCacheMaxSize: Int64 = DEFAULT_DISK_CACHE_MAX_SIZE_BYTES
+
+    /**
+       Enable the slow rendering detection feature.
+     */
+    @objc public var slowRenderingDetectionEnabled: Bool = true
```
seemk, line 152: It was enabled by default due to my mistake. When we have a feature flag it should be disabled by default

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -408,4 +269,22 @@ var splunkRumInitializeCalledTime = Date()
     @objc public class func isInitialized() -> Bool {
         return initialized
     }
+
+    /**
+       Updates the current location. The latitude and longitude will be appended to every span and event.
+     */
+    @objc public class func setLocation(location: CLLocationManager) {
```
seemk, line 276: For ease of use it should be `@objc public class func setLocation(latitude: Double, longitude: Double)`. Users might not want to use CLLocationManager, plus I think it's a bit more clear with explicit latitude and longitude (and reduces coupling).

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -260,10 +119,12 @@ var splunkRumInitializeCalledTime = Date()
         }
         initializeNetworkTypeMonitoring()
         initalizeUIInstrumentation()
-        startSlowFrameDetector(
-                    slowFrameDetectionThresholdMs: options?.slowFrameDetectionThresholdMs,
-                    frozenFrameDetectionThresholdMs: options?.frozenFrameDetectionThresholdMs
-                )
+        if options?.slowRenderingDetectionEnabled ?? true {
```
seemk, line 122: Should default to false as per the other comment for this

seemk (commented):
Was the options building supposed to be part of this PR or was it separate?

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumOptions.swift
Piyush23786: ok

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
Piyush23786: Okay, I'll change CLLocationManager to Double

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
Piyush23786: ok

aurbiztondo-splunk (approved):

PR #101, opened 2022-09-16
APMI 2307: Implements a 15min session timeout

bgustafson:
After the application is in the background for 15min a new rumSessionId will be created upon reopening the app.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
@@ -20,8 +20,7 @@ import UIKit
 import OpenTelemetrySdk
 
 let INACTIVITY_SESSION_TIMEOUT_SECONDS = 15 * 60
-var IS_NEW_SESSION_ID_FOR_INACTIVITY = false
-private var sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+private var sessionIdInactivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
 
 // Constants for lifecyle events that are being observed
 private let UI_APPLICATION_WILL_RESIGN_ACTIVE_NOTIFICATION = "UIApplicationWillResignActiveNotification"
@@ -112,11 +111,10 @@ func lifecycleEvent(_ event: String) {
 func invalidateSession(_ event: String) {
     // 15 min inactivity then session time out
     if event == UI_APPLICATION_WILL_RESIGN_ACTIVE_NOTIFICATION {
-        sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
+        sessionIdInactivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
     } else if event == UI_APPLICATION_WILL_ENTER_FOREGROUND_NOTIFICATION {
-        if Date() > sessionIdInActivityExpiration { // expire 15 min
-           IS_NEW_SESSION_ID_FOR_INACTIVITY = true
-            _  = getRumSessionId()
+        if Date() > sessionIdInactivityExpiration { // expire 15 min
+            _  = getRumSessionId(forceNewSessionId: true)
         }
     }
 }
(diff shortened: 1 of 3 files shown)


bgustafson:
This is a refactor/clean up of this MR. https://github.com/signalfx/splunk-otel-ios/pull/51.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
```diff
@@ -19,6 +19,19 @@ import Foundation
 import UIKit
 import OpenTelemetrySdk
 
+let INACTIVITY_SESSION_TIMEOUT_SECONDS = 15 * 60
+var IS_NEW_SESSION_ID_FOR_INACTIVITY = false
```
johnbley, line 23: I'd much rather see a method like `forceNewSessionId` or similar in `Session.swift` being called from the app lifecycle than this global variable that controls what another file's logic does.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
```diff
@@ -19,6 +19,19 @@ import Foundation
...
+private var sessionIdInActivityExpiration = Date().addingTimeInterval(TimeInterval(INACTIVITY_SESSION_TIMEOUT_SECONDS))
```
johnbley, line 24: Why the odd capitalization of `InActivity` here?

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
```diff
@@ -39,13 +52,13 @@ func initializeAppLifecycleInstrumentation() {
...
      */
     let events = [
-        "UIApplicationWillResignActiveNotification",
-        "UIApplicationSuspendedEventsOnlyNotification",
-        "UIApplicationDidEnterBackgroundNotification",
-        "UIApplicationWillEnterForegroundNotification",
-        "UIApplicationDidBecomeActiveNotification",
-        "UIApplicationSuspendedNotification",
-        "UIApplicationWillTerminateNotification"
+        UI_APPLICATION_WILL_RESIGN_ACTIVE_NOTIFICATION,
```
johnbley, line 55: Good call moving these to constants.

johnbley (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
bgustafson: No reason.  I think that is what the variable was named in the other MR.  I will fix the capitalization.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
bgustafson: 👍

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
bgustafson: What do you think of this?  I added an optional parameter to the getRumSessionId function.  When true gets passed it will run generate a new session id.  If not it will follow the normal path.

johnbley: I like that even better.

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
```diff
@@ -61,6 +61,9 @@ func getRumSessionId() -> String {
         isSessionIdChanged = true
         callbacks = sessionIdCallbacks
     }
+    if forceNewSessionId {
+        rumSessionId = generateNewSessionId()
+    }
```
seemk, line 66: It seems we won't get a session ID change span if a new session ID is forced. Just trying to remember / understand what was the reasoning for this in our ticket. 😄 Feel free to merge if it's not important

seemk (approved):

SplunkRumWorkspace/SplunkRum/SplunkRum/Session.swift
seemk: Asked Seerut about it, she said the session change span is expected

bgustafson: @seemk I removed this if block and if we are forcing a new sessionId then it will go down the same path as when a sessionId expires.

PR #103, opened 2022-10-03
Feature/APMI3255 iOS: options factory update

Piyush23786:
APMI-3255:-
1. Set Location method that sets global attributes (only 2 numbers - latitude and longitude)
2. Bundle identifier (CF Bundle) that is detected and added automatically.
3. Feature flag for slow/frozen renders
4. Contents attribute added.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/Attribute.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/Attribute.swift
@@ -1,0 +1,58 @@
+//
+/*
+Copyright 2021 Splunk Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import Foundation
+
+    enum Attribute {
+
+       static let DEPLOYMENT_ENVIRONMENT      = "environment"
+       static let COMPONENT_KEY               = "component"
+       static let SCREEN_NAME_KEY             = "screen.name"
+       static let LAST_SCREEN_NAME_KEY        = "last.screen.name"
+       static let ERROR_TYPE_KEY              = "exception.type"
+       static let LOCATION_LATITUDE_KEY       = "location.lat"
+       static let LOCATION_LONGITUDE_KEY      = "location.long"
+       static let ERROR_MESSAGE_KEY           = "exception.message"
+       static let ERROR_STACKTRACE_KEY        = "exception.stacktrace"
+       static let RUM_TRACER_NAME             = "SplunkRum"
+
+        static let LINK_TRACE_ID_KEY          = "link.traceId"
+        static let LINK_SPAN_ID_KEY           = "link.spanId"
+
+       static let COMPONENT_APP_START         = "appstart"
+       static let COMPONENT_CRASH             = "crash"
+       static let COMPONENT_ERROR             = "error"
+       static let COMPONENT_UI                = "ui"
+       static let COMPONENT_APP_LIFECYCLE     = "app-lifecycle"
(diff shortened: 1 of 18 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/Attribute.swift
```diff
@@ -0,0 +1,58 @@
...
+       static let SPAN_NAME_APP_ENTERFOREGROUND      = "EnterForeground"
+       static let SPAN_NAME_SCREEN_NAME_CHANGE       = "screen name change"
+       static let SPAN_NAME_APP_START                = "AppStart"
+       static let SPAN_NAME_SESSIONID_CHANGE         = "sessionId.change"
+       static let SPAN_NAME_SHOWVC                   = "ShowVC"
+       static let SPAN_NAME_ACTION                   = "action"
+       static let SPAN_NAME_SPLUNKRUM_INITIALIZE     = "SplunkRum.initialize"
+       static let SPAN_NAME_MANUAL_SPAN              = "manualSpan"
+       static let SPAN_NAME_SLOWRENDERS              = "slowRenders"
+       static let SPAN_NAME_FROZENRENDERS            = "frozenRenders"
```
seemk, line 56: I think these different sections should be separated. Currently there are attribute key names, component attribute value names and span name values grouped together under `Attribute`.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -423,4 +275,19 @@ var splunkRumInitializeCalledTime = Date()
...
+            setGlobalAttributes([Attribute.LOCATION_LONGITUDE_KEY: longitude])
+            setGlobalAttributes([Attribute.LOCATION_LATITUDE_KEY: latitude])
+    }
+
+    /**
+      Set up the deployment environment for RUM instance. This will be passed along as a span
+     */
+    @objc public class func deploymentEnvironment(environment: String) {
+            setGlobalAttributes([Attribute.DEPLOYMENT_ENVIRONMENT: environment])
+    }
```
seemk, line 292: Deployment environment is only set at startup, so this method can be removed

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -423,4 +275,19 @@ var splunkRumInitializeCalledTime = Date()
...
         return initialized
     }
+
+    /**
+       Updates the current location. The latitude and longitude will be appended to every span and event.
+     */
+    @objc public class func setLocation(latitude: Double, longitude: Double) {
+            setGlobalAttributes([Attribute.LOCATION_LONGITUDE_KEY: longitude])
+            setGlobalAttributes([Attribute.LOCATION_LATITUDE_KEY: latitude])
+    }
```
seemk, line 285: 👍 But can do it in a single call to `setGlobalAttributes`

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -24,155 +23,6 @@ import WebKit
...
 
-/**
- Default maximum size of the disk cache in bytes.
- */
-public let DEFAULT_DISK_CACHE_MAX_SIZE_BYTES: Int64 = 25 * 1024 * 1024
-
-/**
- Optional configuration for SplunkRum.initialize()
- */
-@objc public class SplunkRumOptions: NSObject {
```
seemk, line 35: Is the removal from this file transparent to end users? I don't know enough here, but say someone is importing `SplunkRumOptions`, is there a chance the moving of this class will break someone's code?

SplunkRumWorkspace/SplunkRum/SplunkRum/Attribute.swift
Piyush23786: separated as a KeyValue,Component & SpanName.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
Piyush23786: Removed

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
Piyush23786: Done in a single call.

PR #102, opened 2022-09-29
Feature/APMI-3297-Getting an identifier and carrier name directly ios

Piyush23786:
CoreTeleohony framwork added

PR #107, opened 2022-10-26
feat: include carrier information

seemk:
Continuation of https://github.com/signalfx/splunk-otel-ios/pull/102

Changes:
* Add network, carrier info when starting a network span
* Added additional carrier info fields
* Improved thread safety
* Changed connection subtype values to comply with OTel semantic conventions


SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkType.swift
```diff
@@ -15,29 +15,137 @@ See the License for the specific language governing permissions and
...
+        setCarrierInfo(telephonyNetworkInfo, identifier: telephonyNetworkInfo.serviceCurrentRadioAccessTechnology?.keys.first)
+        telephonyNetworkInfo.serviceSubscriberCellularProvidersDidUpdateNotifier = { identifier in
+            setCarrierInfo(telephonyNetworkInfo, identifier: identifier)
+        }
+    }
+}
+
+func getNetworkInfo() -> NetworkInfo {
+    pthread_rwlock_rdlock(&netInfoLock)
+    var info = currentNetInfo
```
johnbley, line 131: This should be a local copy (under lock) that gets returned- `var info = NetworkInfo(currentNetInfo)` with associated copy constructor.  Otherwise the threading looks good.  Haven't really looked at the build failure.

seemk: As far as I've understood, structs are copied each time in Swift

Fixed the compilation issue, `CTTelephonyNetworkInfo` wasn't available on mac catalyst. Added an `#if os(iOS)` around the code and on mac it will return an empty struct

johnbley: Whoops, my mistake.  That's what you get for switching languages too quickly.  Code looks fine.

PR #108, opened 2022-10-27
fix: clear carrier info on non-cell connection types

seemk:
Switching from cellular to wifi did not clear carrier information.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkType.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkType.swift
@@ -148,6 +148,8 @@ func initializeNetworkTypeMonitoring() {
                 } else {
                     setConnectionType(nil)
                 }
+            } else {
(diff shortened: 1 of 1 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkType.swift
```diff
@@ -43,6 +43,13 @@ private func setConnectionType(_ type: String?) {
         pthread_rwlock_unlock(&netInfoLock)
     }
     currentNetInfo.hostConnectionType = type
+
+    if type != "wifi" {
```
bgustafson, line 47: Should this if type == "wifi"

seemk: Changed it to `type != `cell` since it might be possible it is nil (no idea in which case though)

bgustafson (approved):

jnino-splunk (approved):

PR #109, opened 2022-10-28
fix:switching to a wifi network, the carrier info should be cleared

Piyush23786:
switching to a wifi network, the carrier info should be cleared.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkType.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkType.swift
@@ -43,7 +43,6 @@ private func setConnectionType(_ type: String?) {
         pthread_rwlock_unlock(&netInfoLock)
     }
     currentNetInfo.hostConnectionType = type
-
 }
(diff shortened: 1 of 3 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkType.swift
```diff
@@ -44,12 +44,6 @@ private func setConnectionType(_ type: String?) {
     }
     currentNetInfo.hostConnectionType = type
 
-    if type != "cell" {
-        currentNetInfo.hostConnectionSubType = nil
-        currentNetInfo.carrierCountryCode = nil
-        currentNetInfo.carrierNetworkCode = nil
-        currentNetInfo.carrierIsoCountryCode = nil
-    }
```
seemk, line 52: Shouldn't this part stay though? `setCarrierInfo` is only called from `serviceSubscriberCellularProvidersDidUpdateNotifier` callback, is this callback invoked when the connection type changes?

PR #114, opened 2022-11-15
APMI-2771 -App startup timing improvements

jnino-splunk:
This PR adds additional functionality to the app startup timing. In particular it attempts to catch 3 scenarios were the app startup is likely to be incorrect as "the user's perceived startup". 

This checks for "prewarming" - a feature in iOS 15+ that can start setting the application up before the user launches the app.

This also checks for instances where the app was in a background state before we get a didBecomeActive notification.

This also catches timing that is well beyond a reasonable threshold (5 mins). 

This also just does not report in these cases as I believe it is better to have consistency with timing logic.

I added code comments where appropriate.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,30 +6,31 @@ let package = Package(
     name: "SplunkOtel",
     platforms: [
         .iOS(.v11),
-	.macOS(.v10_13)
+        .macOS(.v10_13)
     ],
     products: [
         .library(name: "SplunkOtel", targets: ["SplunkOtel"])
     ],
     dependencies: [
         .package(name: "opentelemetry-swift", url:"https://github.com/open-telemetry/opentelemetry-swift", from: "1.1.4"),
-	.package(url: "https://github.com/devicekit/DeviceKit.git", from: "4.0.0"),
+        .package(url: "https://github.com/devicekit/DeviceKit.git", from: "4.0.0"),
     ],
     targets: [
         .target(
             name: "SplunkOtel",
             dependencies: [
-		.product(name: "OpenTelemetryApi", package:"opentelemetry-swift"),
-		.product(name: "OpenTelemetrySdk", package:"opentelemetry-swift"),
-		.product(name: "StdoutExporter", package:"opentelemetry-swift"),
-		.product(name: "ZipkinExporter", package:"opentelemetry-swift"),
-		.product(name: "DeviceKit", package: "DeviceKit")
+                .product(name: "OpenTelemetryApi", package:"opentelemetry-swift"),
+                .product(name: "OpenTelemetrySdk", package:"opentelemetry-swift"),
+                .product(name: "StdoutExporter", package:"opentelemetry-swift"),
+                .product(name: "ZipkinExporter", package:"opentelemetry-swift"),
+                .product(name: "DeviceKit", package: "DeviceKit")
             ],
             path: "SplunkRumWorkspace/SplunkRum",
             exclude: [
-		"SplunkRumTests",
-		"SplunkRum/SplunkRum.h",
-		"SplunkRum/Info.plist"
+                "SplunkRumTests",
+                "SplunkRumDiskExportTests",
+                "SplunkRum/SplunkRum.h",
+                "SplunkRum/Info.plist"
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -6,30 +6,31 @@ let package = Package(
...
+                .product(name: "ZipkinExporter", package:"opentelemetry-swift"),
+                .product(name: "DeviceKit", package: "DeviceKit")
             ],
             path: "SplunkRumWorkspace/SplunkRum",
             exclude: [
-		"SplunkRumTests",
-		"SplunkRum/SplunkRum.h",
-		"SplunkRum/Info.plist"
+                "SplunkRumTests",
+                "SplunkRumDiskExportTests",
```
jnino-splunk, line 31: I re-indented the whole file. But more importantly, I also added SplunkRumDiskExportTests to the exclude list. This get rid of a package warning that those files were not properly handled.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppLifecycle.swift
```diff
@@ -96,7 +96,7 @@ func lifecycleEvent(_ event: String) {
     if event == UI_APPLICATION_WILL_TERMINATE_NOTIFICATION {
         let now = Date()
         let span = buildTracer().spanBuilder(spanName: "AppTerminating").setStartTime(time: now).startSpan()
-        span.setAttribute(key: "component", value: "AppLifecycle")
+        span.setAttribute(key: "component", value: "app-lifecycle")
```
jnino-splunk, line 99: above on line 78, this is set as "app-lifecycle". I thought they should match. Given the Spans test to be capitalized camel case, I used the style from line 78 for the component.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppStart.swift
```diff
@@ -42,39 +42,101 @@ private func processStartTime() throws -> Date {
...
-                appStart!.addEvent(name: event.rawValue)
-                if event == UIApplication.didBecomeActiveNotification {
-                    appStart!.end()
-                    OpenTelemetry.instance.contextProvider.removeContextForSpan(appStart!)
-                    appStart = nil
-                    // Because of heavy overlap and desired treatment of AppStart vs
-                    // ongoing app lifecycle stuff, initialize this now rather than
-                    // earlier to avoid double-reporting or more complex logic
-                    initializeAppLifecycleInstrumentation()
-                }
```
jnino-splunk, line 70: I switched this up to use a closure for each so that we can remove the observers after we get the notification. Before these would stay alive for the whole session.

SplunkRumWorkspace/SplunkRum/SplunkRum/AppStart.swift
```diff
@@ -42,39 +42,101 @@ private func processStartTime() throws -> Date {
...
+        // the app was made active from the background,
+        // or we are over a nonsensical threshold, we do not report the appStart span.
+        if (isPrewarm && prewarmAvailable)
+            || wasBackgroundedBeforeWillEnterForeground
+            || (Date().timeIntervalSince1970 - (spanStart.timeIntervalSince1970)) > possibleAppStartTimingErrorThreshold {
+            OpenTelemetry.instance.contextProvider.removeContextForSpan(appStart)
+        } else {
+            appStart.addEvent(name: notification.name.rawValue)
+            appStart.end()
+            OpenTelemetry.instance.contextProvider.removeContextForSpan(appStart)
```
seemk, line 112: I think calling `end` should already remove it: https://github.com/open-telemetry/opentelemetry-swift/blob/main/Sources/OpenTelemetrySdk/Trace/RecordEventsReadableSpan.swift#L281

jnino-splunk: Oh interesting. Will update.

seemk (approved):

johnbley (approved):
I might consider adding more span fields to distinguish some of these additional cases (vague idea I know), but overall LGTM.

PR #116, opened 2022-11-29
chore: disable cyclomatic_complexity lint rule

seemk:
CI had it's swiftlint upgraded, which started failing on the current codebase due to `orphaned_doc_comment` rule (https://github.com/signalfx/splunk-otel-ios/actions/runs/3477866164/jobs/6009506623#step:3:40).

This rule was triggered when there was a doc comment before a function that had a `swiftlint:disable:next` clause for a different rule. E.g. `SplunkRum.initialize` had `//swiftlint:disable:next cyclomatic_complexity`.

Added `cyclomatic_complexity` to ignored rules, don't think it's useful anyway.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -10,6 +10,7 @@ disabled_rules:
  - function_body_length
  - force_cast
  - private_over_fileprivate
+ - cyclomatic_complexity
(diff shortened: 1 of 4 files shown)


johnbley (approved):
Good call

PR #120, opened 2022-12-07
fix: unhandled promise rejection in webview instrumentation

seemk:
It seems `postMessage` required at least 1 argument (https://developer.apple.com/documentation/webkit/wkscriptmessagehandler).

Calling it without arguments returned a rejecting promise, which was picked up by the Web instrumentation and reported as an error.

Added the extra promise catch just in case, even though adding the single parameter fixed the issue.


johnbley (approved):
Nice!

PR #123, opened 2022-12-29
NetworkInstrumentation without KVO

Piyush23786:
NetworkInstrumentation without KVO

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkInstrumentation.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkInstrumentation.swift
@@ -121,115 +121,109 @@ func startHttpSpan(request: URLRequest?) -> Span? {
     return span
 }
 
-class SessionTaskObserver: NSObject {
+class SessionTask: NSObject {
     var span: Span?
     // Observers aren't kept alive by observing...
-    var extraRefToSelf: SessionTaskObserver?
+    var extraRefToSelf: SessionTask?
     var lock: NSLock = NSLock()
     override init() {
         super.init()
         extraRefToSelf = self
     }
 
-    override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey: Any]?, context: UnsafeMutableRawPointer?) {
-        lock.lock()
-        defer {
-            lock.unlock()
-        }
-        let task = object as? URLSessionTask
-        if task == nil {
-            return
-        }
-        if span == nil {
-            span = startHttpSpan(request: task!.originalRequest)
-        }
-        // FIXME possibly also allow .canceling to close the span?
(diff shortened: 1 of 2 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkInstrumentation.swift
```diff
@@ -121,115 +121,109 @@ func startHttpSpan(request: URLRequest?) -> Span? {
...
-            span = startHttpSpan(request: task!.originalRequest)
-        }
-        // FIXME possibly also allow .canceling to close the span?
-        if task!.state == .completed && extraRefToSelf != nil {
-            endHttpSpan(span: span,
-                        task: task!)
-            task!.removeObserver(self, forKeyPath: "state")
-            extraRefToSelf = nil
+    func networkSpan(task: URLSessionTask) {
+        DispatchQueue.main.async {
```
seemk, line 135: Is the async queue needed here?

SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkInstrumentation.swift
```diff
@@ -121,115 +121,109 @@ func startHttpSpan(request: URLRequest?) -> Span? {
...
+            if self.span == nil {
+                self.span = startHttpSpan(request: task.originalRequest)
+            }
+
+            if  self.extraRefToSelf != nil {
```
seemk, line 140: Are you sure it's not calling `endHttpSpan` right away after starting a span here? Have you checked the span durations? Since `extraRefToSelf` is assigned in init, this check is always true and the span is ended prematurely.

Besides, without KVO `extraRefToSelf` should no longer be necessary.

seemk (changes requested):

SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkInstrumentation.swift
Piyush23786: In order to create a span, async is required.

SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkInstrumentation.swift
Piyush23786: Yes, I checked the span durations. then according to the response, endHttpSpan should be called?

SplunkRumWorkspace/SplunkRum/SplunkRum/NetworkInstrumentation.swift
seemk: Why?

seemk:
Going with https://github.com/signalfx/splunk-otel-ios/pull/124 at the moment

PR #131, opened 2023-02-10
change scheme/target name to SplunkOtel

seemk:
Fixes inconsistencies where the module was imported as `SplunkRum` in the local workspace and `SplunkOtel` otherwise.

Due to https://github.com/apple/swift/issues/56573 we can't currently create an XCFramework because the module name was the same as an exported class (`SplunkRum`), changing the module name to `SplunkOtel` fixes this.


johnbley (approved):
Yep, this was an overdue refactoring from when I did the package rename way back when.

PR #134, opened 2023-03-03
Pinning open telemetry swift to exact 1.2.0 for breaking change in 1.4.0

jnino-splunk:
Pins the open-telemetry-swift version to 1.2.0 so that 1.4.0 is not pulled in which has a breaking change.


jnino-splunk:
@johnbley  I wasn't sure on the process here but I created a new branch off of v.10.0 and then created my branch with the change

jnino-splunk:
I tried 1.3.0 and 1.2.1 and both caused test failures 😬

PR #137, opened 2023-04-16
Add SauceLabs tests to CI

seemk:
There's now a new project, SauceLabsTestApp, which runs through a few (mostly) networking tests. The app sets up its own HTTP server which is then used to route the spans to tests for verification.

I tried to bring in some order to the tests, all tests derive from `TestCase`, which sets up some central bookkeeping (handles timeouts, attaches test names to spans, routes test results).

When creating new tests, only 2 methods need to be overridden: `execute` - sets up the test, e.g. start an HTTP request and `verify` which is called with the received spans.

Since it's a bit annoying to verify the results in Sauce Labs, I just created a label that changes the text to `success` when all tests pass. Sauce Labs waits for this label.

Areas that need more work:
* iOS 11-13 compatibility. Initially opted for SwiftUI for the test app, but later found out it's not really supported on iOS 11 and 12. The UI part of the app is really small, so perhaps it would make sense to write it in the old way. **iOS 13**, the app compiles and tests run fine, but the Sauce Labs script can't find the UI elements necessary to validate tests.
* More tests (crash reporting, additional networking libraries, UI clicks)?


SplunkRumWorkspace/SauceLabsTestApp/sl_tests.py
```diff
@@ -0,0 +1,45 @@
...
+class IOSTests(unittest.TestCase):
+
+    def setUp(self):
+        currentTime = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+        caps = {}
+
+        sl_file_id = sys.argv[2]
+        
+        caps['platformName'] = 'iOS'
+        caps['appium:app'] = f'storage:{sl_file_id}'
```
mhennoch, line 19: Typo?
caps['appium:app'] = f'storage:{sl_file_id}'

.github/workflows/ci.yml
```diff
@@ -31,6 +31,62 @@ jobs:
...
+          echo "sl_file_id=$item_id" >> $GITHUB_OUTPUT
+    outputs:
+      sl_file_id: ${{ steps.sl-upload.outputs.sl_file_id }}
+  saucelabs-tests:
+    runs-on: macOS-latest
+    needs: saucelabs-prepare
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['14', '15', '16']
```
johnbley, line 73: This is great.  I really like that the test matrix is in github rather than in SauceLabs UI/config.

johnbley (approved):

seemk:
I have read the CLA Document and I hereby sign the CLA

PR #138, opened 2023-05-11
MOB-3112: Options Factory Update.

bgustafson:
- Created a SplunkRumBuilder class that abstracts the SplunkRumOptions away from users.  This closely matches how the Android SDK is initialized.
- Created a constants struct so that there is a central location for attribute, span and event names.


SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
```diff
@@ -0,0 +1,144 @@
...
+    @discardableResult
+    @objc
+    public func allowInsecureBeacon(enabled: Bool) -> SplunkRumBuilder {
+        self.allowInsecureBeacon = enabled
+        return self
+    }
+
+    @discardableResult
+    @objc
+    public func debugEnabled(enabled: Bool) -> SplunkRumBuilder {
```
seemk, line 51: Maybe a question of personal taste, but perhaps it would make sense to skip the double `enabled` wording by omitting the parameter name? I.e. `.debugEnabled(enabled: true)` vs `.debugEnabled(true)`

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -193,7 +198,7 @@ var splunkRumInitializeCalledTime = Date()
      
      */
     @discardableResult
-    @objc public class func initialize(beaconUrl: String, rumAuth: String, options: SplunkRumOptions? = nil) -> Bool {
+    @objc internal class func initialize(beaconUrl: String, rumAuth: String, options: SplunkRumOptions? = nil) -> Bool {
```
seemk, line 201: Since moving to the new pattern is a breaking change, perhaps we should mark this function along with the previous Options (if it's possible for non-funcs) deprecated? And only after a while remove it completely

bgustafson: Yeah that makes sense.  I can revert this and mark this and the SplunkRumOptions as deprecated since the options are not going to be exposed in a future release.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
bgustafson: I am not sure that an unnamed parameter would work here since it needs to be used in the enclosed function.  I can change the signature to "debug(enabled: Bool)" to shorten the name and remove the duplicate "enabled"

seemk (approved):

README.md
```diff
@@ -31,7 +31,9 @@ this is probably your `AppDelegate`'s `...didFinishLaunchingWithOptions:` method
 import SplunkOtel
 ...
 // Your beaconUrl and rumAuth will be provided by your friendly Splunk representative
-SplunkRum.initialize(beaconUrl: "https://rum-ingest.us0.signalfx.com/v1/rum", rumAuth: "ABCD...")
+SplunkRumBuilder(beaconUrl: "https://rum-ingest.us0.signalfx.com/v1/rum", rumAuth: "ABCD...")
+  // Call functions to configure additional options
+  .build()
```
theletterf, line 36: How do you add the environment here? Would this be accurate? We should add this to the official docs:

```swift
import SplunkOtel
//..
SplunkRumBuilder(beaconUrl: "https://rum-ingest.<realm>.signalfx.com/v1/rum", rumAuth: "<rum-token>")
// Call functions to configure additional options
   .globalAttributes(globalAttributes: ["deployment.environment": "<environment>"])
    .build()
```

README.md
```diff
@@ -41,7 +43,9 @@ or
 @import SplunkOtel;
 ...
 // Your beaconUrl and rumAuth will be provided by your friendly Splunk representative
-[SplunkRum initializeWithBeaconUrl: @"https://rum-ingest.us0.signalfx.com/v1/rum" rumAuth: @"ABCD..." options: nil];
+SplunkRumBuilder *builder = [SplunkRumBuilder withBeaconUrl: @"https://rum-ingest.us0.signalfx.com/v1/rum" rumAuth: @"ABCD..."];
+// Call functions to configure additional options
+[builder build];
```
theletterf, line 48: What would the ObjC equivalent of setting a global attribute in Swift?

Our previous code was:

```objc
@import SplunkOtel;

//...
SplunkRumOptions *options = [[SplunkRumOptions alloc] init];
options.environment = @"<environment-name>";
[SplunkRum initializeWithBeaconUrl:@"https://rum-ingest.<realm>.signalfx.com/v1/rum" rumAuth: @"<rum-token>" options: options];
```

README.md
```diff
@@ -41,7 +43,9 @@ or
...
 ```
 
 ## Installation options
```
theletterf, line 51: How do you set these options now? I'm guessing:

**Swift**

```swift
import SplunkOtel
import SplunkOtelCrashReporting
//..
SplunkRum.initialize(beaconUrl: "https://rum-ingest.<realm>.signalfx.com/v1/rum",
    rumAuth: "<rum-token>",
    options: SplunkRumOptions(environment:"<environment-name>"))
     // Initialize crash reporting module after the iOS agent
SplunkRumCrashReporting.start()
```

**Objective-C**

```objc
@import SplunkOtel;
@import SplunkOtelCrashReporting;
//...
SplunkRumBuilder *builder = [SplunkRumBuilder withBeaconUrl: @"https://rum-ingest.<realm>.signalfx.com/v1/rum" rumAuth: @"<rum-token>"];
// Call functions to configure additional options
     [builder build];
// Initialize crash reporting module after the iOS agent
[SplunkRumCrashReporting start]
```

SplunkRumWorkspace/SplunkRum/SplunkRum/Constants.swift
```diff
@@ -0,0 +1,78 @@
...
+    struct SpanNames {
+        static let SPLUNK_RUM_INITIALIZE = "SplunkRum.initialize"
+        static let SESSION_ID_CHANGE = "sessionId.change"
+        static let APP_START = "AppStart"
+        static let APP_TERMINATING = "AppTerminating"
+        static let RESIGNACTIVE = "ResignActive"
+        static let ENTER_FOREGROUND = "EnterForeground"
+        static let ACTION = "action"
+        static let PRESENTATION_TRANSITION = "PresentationTransition"
+        static let SCREEN_NAME_CHANGE = "screen name change"
```
theletterf, line 34: Are the following functions still available? Do they work the same?

```swift
   // You can also call the ``setGlobalAttributes`` function 
   // anywhere in your code using the same map
   SplunkRum.setGlobalAttributes([])
   
   // To remove a global attribute, pass the key name to removeGlobalAttribute
   SplunkRum.removeGlobalAttribute("key2")

   SplunkRum.setScreenName("AccountSettingsTab")

   SplunkRum.reportError(example_error)
```

theletterf (changes requested):
We can't go forward until we've cleared missing information for the docs.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
```diff
@@ -0,0 +1,144 @@
...
+    @discardableResult
+    @objc
+    public func globalAttributes(globalAttributes: [String: Any]) -> SplunkRumBuilder {
+        self.globalAttributes = globalAttributes
+        return self
+    }
+
+    @discardableResult
+    @objc
+    public func environment(environment: String) -> SplunkRumBuilder {
```
seemk, line 65: To keep consistency with `deployment.environment`, perhaps `deploymentEnvironment`?

theletterf: Why not letting users set a global attribute? Or it's for backward compatibility reasons?

theletterf: Is there a preferred way going forward that we can promote above the others?

README.md
bgustafson:
```swift
SplunkRumBuilder(beaconUrl: "https://rum-ingest.<realm>.signalfx.com/v1/rum", rumAuth: "<rum-token>")
   .globalAttributes(globalAttributes: [:])
  .deploymentEnvironment(environment: "<environment>")
   .build()
```

README.md
bgustafson:
```objc
@import SplunkOtel;

SplunkRumBuilder *builder = [SplunkRumBuilder withBeaconUrl: @"https://rum-ingest.us0.signalfx.com/v1/rum" rumAuth: @"ABCD..."];
[builder withGlobalAttributes: [:]]
[builder build];
```

README.md
theletterf: The ObjC snippet is OK?

bgustafson: Yes, the ObjC snippet is okay because it is not using the builder not SplunkRum initialize.

README.md
theletterf: Would you still do `options.environment = @"<environment-name>";`?

README.md
theletterf: What about settings like `screenNameSpans` or any other in this list? https://github.com/signalfx/splunk-otel-ios#installation-options

We need to explain how those are set for Swift and ObjC.

bgustafson: They all are set on the builder.  For screenNameSpans it would look like this

```swift
builder.screenNameSpans(enabled: true)
```
and the above needs to be called before builder.build().

I see all the options listed in the readMe but no example on how to set them in the options.  All the defaults have remained the same.  Do you want an example like below except expanded to show each option that could be changed?

```swift
SplunkRumBuilder(beaconUrl: "http://127.0.0.1:8989/v1/traces", rumAuth: "FAKE")
        .allowInsecureBeacon(enabled: true)
        .debug(enabled: true)
        .globalAttributes(globalAttributes: ["strKey": "strVal", "intKey": 7, "doubleKey": 1.5, "boolKey": true])
        .deploymentEnvironment(environment: "env")
        .ignoreURLs(ignoreURLs: try! NSRegularExpression(pattern: ".*ignore_this.*"))
        .screenNameSpans(enabled: true)
        .build()
```

theletterf: I think that'll suffice—thanks! :)

theletterf (approved):
Only two remaining questions are about ObjC unfortunately...

- How do you configure settings in ObjC now? I suspect it would be a bunch of `[builder sampleSetting: value] `statements? Not very familiar with ObjC syntax.
- Is there still a direct setting for environment accessible by ObjC? What's the syntax? `[builder environment: "environment-name"] `?

bgustafson:
Your assumption is correct about the ObjC syntax it will be a bunch of `[builder sampleSetting: value]`.  Specifically for environment it would look like 

```objc 
[builder deploymentEnvironment: @"environment-name"];
```

SplunkRumWorkspace/SplunkRum/SplunkRum/Constants.swift
```diff
@@ -0,0 +1,78 @@
...
+
+import Foundation
+
+struct Constants {
+    struct Globals {
+        static let INSTRUMENTATION_NAME = "splunk-ios"
+        static let UNKNOWN_APP_NAME = "unknown-app"
+    }
+    struct SpanNames {
+        static let SPLUNK_RUM_INITIALIZE = "SplunkRum.initialize"
```
theletterf, line 26: We write in the docs:

```
   * - ``thread.name``
     - String
     - Name of the thread. ``SplunkRum.initialize`` only applies to the main thread.
```

I take this is unchanged despite the renaming of the initialization function?

bgustafson: Yes, this should be unchanged because we are still initializing the SplunkRum object on this thread.

SplunkRumWorkspace/SplunkRum/SplunkRum/Constants.swift
```diff
@@ -0,0 +1,78 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import Foundation
+
+struct Constants {
```
jnino-splunk, line 20: Just a comment here. I don't necessarily think it needs a change, but I think it is better to use an enum over a struct when the purpose is to namespace some group of constants like this. It makes it so it is impossible to ever instantiate the type `Constants`. A enum can have no cases and just serve as a namespace!

jnino-splunk: a private init would also work, of course 👍

theletterf:
@bgustafson Will there be a `realm` setting, as per https://github.com/signalfx/gdi-specification/blob/65af46e25075cf910038b971d72b55d5555e4cc0/specification/configuration.md?plain=1#L270 ?

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -288,6 +296,104 @@ var splunkRumInitializeCalledTime = Date()
         initialized = true
         print("SplunkRum.initialize() complete")
         return true
+    }
+
+    @discardableResult
+    @objc internal class func create(beaconUrl: String, rumAuth: String, options: SplunkRumOptions? = nil) -> Bool {
```
jnino-splunk, line 302: should we just make the options non-optional here? This is only called from the builder correct? And that always passes non nil options.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
```diff
@@ -0,0 +1,149 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import Foundation
+
+@objc public class SplunkRumBuilder: NSObject {
```
jnino-splunk, line 20: Will we be documenting these build funcs with some description of what they do?

jnino-splunk (commented):
a few things to perhaps look at before merging.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
theletterf: Like dev docs? Those can live in the repo—if they're for developing the distro.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
theletterf: The ones that should be mandatory, in my opinion, are:

- beaconUrl || realm
- rumAuth

Rest could be optional. But I'd like to see realm coming first, to replicate the same behavior of our other RUM instrumentations.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
jnino-splunk: Right like the dev docs seen here on the android builder https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/SplunkRumBuilder.java

I think it is useful for users of the lib to see what the function / property does right in the IDE

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
jnino-splunk: makes sense! I just think since the builder always calls into this function with a non-nil SplunkOptions object, it would clean things up here to not accept an optional. The user can of course chose to call `build` with as many builder configuration functions as they need :)

jnino-splunk: Also, Android has a requirement on applicationName. Should iOS also introduce this requirement and add it to the init params?

theletterf: Not if the iOS distro already retrieves that name automatically.

jnino-splunk: it does but it can fail. And that name might not be what the user wants to see on their dashboard. Basically, should iOS and Android be the same here?

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
theletterf: I don't know. We need a RUM design overlord to decide. :)

jnino-splunk: Btw. All my comments are in relation to a task I am working on - document differences between iOS and Android. Thanks for your input.

dvernon-splunk:
Picking this up from Brian.
* Bumped the Podspec file to align it with the version bump in this MR
* Agreed with Juan on the change to enums. It was a simple enough change and ensures no one mistakenly instantiates a namespace.
* Also agree with Juan on making SplunkRumOptions non-optional in the create function. My biggest worry with allowing nil options is that it makes it more difficult for us to control whether an option is true or false by default. The SplunkRumOptions file should imo be where those decisions live, not in the create function or potentially scattered throughout the code.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
theletterf: I think it's best if we replicate behavior as much as possible. +CC @dvernon-splunk

jnino-splunk: pending a possible change here, the new commits look fine to me

dvernon-splunk: Added an optional appName parameter to the create function. Not providing an appName will fall back on old behavior.

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRumBuilder.swift
```diff
@@ -132,18 +132,19 @@ import Foundation
...
-                                                   screenNameSpans: self.screenNameSpans,
-                                                   networkInstrumentation: self.networkInstrumentation,
-                                                   enableDiskCache: self.enableDiskCache,
-                                                   spanDiskCacheMaxSize: self.spanDiskCacheMaxSize,
-                                                   slowFrameDetectionThresholdMs: self.slowFrameDetectionThresholdMs,
-                                                   frozenFrameDetectionThresholdMs: self.frozenFrameDetectionThresholdMs,
-                                                   sessionSamplingRatio: self.sessionSamplingRatio))
+                                rumAuth: self.rumAuth,
+                                appName: nil,
+                                options: .init(allowInsecureBeacon: self.allowInsecureBeacon,
```
jnino-splunk, line 137: where would be set to something other than nil?

dvernon-splunk: Including setApplicationName in builder.

jnino-splunk (approved):

PR #147, opened 2023-06-09
MOB-3104: faqs doc

jnino-splunk:
A doc of FAQs for the repo


jnino-splunk:
@bgustafson @akubik-splunk Here is a draft of the faqs doc. I am guessing this will not live in the repo similar to the android one @theletterf?

theletterf:
@jnino-splunk That's right. Plus, we don't really do FAQs in our docs.

I'll integrate all these items in the in-progress docs PR. This will be the start of the iOS troubleshooting doc.

PR #142, opened 2023-05-22
MOB-3105 adds podspec file

dvernon-splunk:
Adds the podspec file to support cocoapod distribution. As our readme states we're still in beta, I started versioning at 0.1.0, but can start at 1.0.0 if that's preferred. Once the podspec is merged into main, I'll publish the pod for use.

PR #150, opened 2023-07-06
MOB-3105 Creates Podspec File for Cocoapods Support

dvernon-splunk:
Same as https://github.com/signalfx/splunk-otel-ios/pull/142 but remade as I had to mess with my SSH keys and it unverified my commits.

Creates the podspec file for cocoapods support.
Adds a version check in the fullbuild script to ensure the version number on the Swift Package and Cocoapod is the same. 
Removes a broken lint check in the UI test that was preventing the build from completing.

### DIFF
--- a/fullbuild.sh
+++ b/fullbuild.sh
@@ -3,7 +3,7 @@ set -ex
 swiftlint --strict
 
 # Make sure the version numbers on the podspec and SplunkRum.swift match
-echo "Check 3"
+echo "Checking that version numbers match"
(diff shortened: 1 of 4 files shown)


fullbuild.sh
```diff
@@ -1,6 +1,19 @@
 #!/bin/bash
 set -ex
 swiftlint --strict
+
+# Make sure the version numbers on the podspec and SplunkRum.swift match
+echo "Check 3"
```
johnbley, line 6: `Checking that version numbers match` or similar would be a better log line than `Check 3`

johnbley (approved):

PR #154, opened 2023-08-01
Dvernon: MOB-3260/MOB-3261 - New Options

dvernon-splunk:
Creates new options for:
* slowRenderingDetectionEnabled - allows user ability to disable slowRenderingDetection
* spanScheduleDelay - allows user to set their own schedule delay time interval for the BatchSpanProcessor

@theletterf the slowRenderingDetectionEnabled is already mentioned in the docs, but it claims that the default value is false. The default value should be changed to true, as that was the previous default behavior (the customer didn't have a way to turn it off) and it's Android's default behavior as well.

### DIFF
--- a/SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
+++ b/SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
@@ -51,7 +51,8 @@ public let DEFAULT_DISK_CACHE_MAX_SIZE_BYTES: Int64 = 25 * 1024 * 1024
                       slowRenderingDetectionEnabled: Bool = true,
                       slowFrameDetectionThresholdMs: Double = 16.7,
                       frozenFrameDetectionThresholdMs: Double = 700,
-                      sessionSamplingRatio: Double = 1.0
+                      sessionSamplingRatio: Double = 1.0,
+                      spanSchedulingDelay: TimeInterval = 5.0
     ) {
         // rejectionFilter not specified to make it possible to call from objc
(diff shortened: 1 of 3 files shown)


SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -156,6 +166,11 @@ public let DEFAULT_DISK_CACHE_MAX_SIZE_BYTES: Int64 = 25 * 1024 * 1024
      */
     @objc public var sessionSamplingRatio: Double = 1.0
 
+    /**
+     Set the time interval between batch span sampling in seconds
```
seemk, line 170: Perhaps a more apt description would be "maximum interval between 2 consecutive span exports"

SplunkRumWorkspace/SplunkRum/SplunkRum/SplunkRum.swift
```diff
@@ -156,6 +166,11 @@ public let DEFAULT_DISK_CACHE_MAX_SIZE_BYTES: Int64 = 25 * 1024 * 1024
...
+     */
+    @objc public var spanSchedulingDelay: TimeInterval = 5.0
```
seemk, line 172: A few alternative naming ideas: `bspScheduleDelay`, `batchSpanProcessorScheduleDelay` to hint that it's related to OTel's batch span processor? For example in the OTel spec the environment variable corresponding to this option is named `OTEL_BSP_SCHEDULE_DELAY`

seemk (approved):

PR #162, opened 2023-08-18
Dvernon/MOB-3294 Parsing traceparent

dvernon-splunk:
Fixes the parse on reading the traceparent so it will work even if "traceparent" isn't the first pattern in the string. Also cleaned up the syntax to make it a bit more "swifty" and remove some nested ifs.


seemk (commented):
Thanks! Could we add a test case to https://github.com/signalfx/splunk-otel-ios/blob/main/SplunkRumWorkspace/SplunkRum/SplunkRumTests/ServerTimingTests.swift#L72?

E.g.

```
foo; bar=42, traceparent;desc="00-b4cfffe039d464890cf09c1dd766a130-9994088852321abc-01"
foo; bar=42, traceparent;desc="00-b4cfffe039d464890cf09c1dd766a130-9994088852321abc-01", origin; dur=1002
```

PR #172, opened 2023-09-28
Add Traceparent to Outgoing Requests

dvernon-splunk:
Adds the traceparent to the header of outgoing HTTP requests.

Unfortunately, the request object is get-only where we were creating spans, requiring us to swizzle the task functions in URLSessionTask. This solution was adapted, and heavily modified, from opentelemetry-swift: https://github.com/open-telemetry/opentelemetry-swift/blob/8d6fa745c186e3b7556f330b5236da25fa7c9d62/Sources/Instrumentation/URLSession/URLSessionInstrumentation.swift 

The proposed solution swizzles urlsession tasks that take in (or requests) a request object, and passes through a modified request object with the added traceparent header.

Also includes a new variable in SplunkRumOptions, allowing users to enable. It is false by default.


dvernon-splunk:
Found an issue while debugging the failing unit test: Do not review yet.

dvernon-splunk:
The unit tests are still failing, but I wanted to get this proposed solution in before I leave for vacation. The data task unit test succeeds if you run it by itself. If this direction is something we're good with, I will fix the unit tests (and bugs, and lint errors) when I return.

PR #177, opened 2023-10-20
Pin GH actions and runners to the exact versions

Kielek:


.github/workflows/ci.yml
```diff
@@ -6,10 +6,10 @@ on:
 
 jobs:
   full-build:
-    runs-on: macOS-latest
+    runs-on: macOS-12
```
Kielek, line 9: Lower in the file, you are using macos-11, please check what is appropriate version.

dvernon-splunk: Upgrading to 12 should be fine

PR #185, opened 2023-11-24
sampling: use session id for source of randomness

seemk:
Use the session ID bytes for the source of randomness when making a sampling decision.

The algorithm for the session ID value calculation is the same as is in the [JS RUM SDK](https://github.com/signalfx/splunk-otel-js-web/blob/main/packages/web/src/SessionBasedSampler.ts#L88). This fixes the issue where, if sampling is used with the iOS SDK and inside a WebView, the SDKs come to a different sampling decision due to differing algorithms.

Note: session ID callbacks should have the session ID as a parameter, but this would be a breaking change in the public API, so I decided to keep the plain callbacks for now.


johnbley (approved):
Have you checked android?

seemk:
> Have you checked android?

Android has the same issue. The lower 8 bytes of the session ID are converted to a number, this differs from the JS RUM implementation which xors the whole 16 bytes in 4 byte chunks.

PR #186, opened 2023-11-27
chore: update changelog / release 0.11.3

seemk:
Looks like the podfile and SplunkRum.swift are already at 0.11.3


seemk:
Only check_links failing due to the yet-non-existent tag in the changelog update

johnbley (approved):
lgtm

PR #209, opened 2024-04-22
add privacy manifest

seemk:
As required per https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests

I am not sure anything else is applicable in this list.


seemk:
Should be fixed now and added the privacy info to the podspec as well

PR #266, opened 2025-04-02
DEMRUM-1403: Update swift API

jCharv:
This PR updates the swift API. The update is not yet as per spec, as spec changes were made during the implementation, and further updates will come later (mainly the endpoint configuration).

This PR also updates Resources, sets up Traces and Logs exporters, does some changes to the Network instrumentation (so that the agent is not instrumenting itself) to be able to test e2e.


vaclav-halik-smartlook (changes requested):
Looks good. But there are still a few minor things that need to be tweaked. Details sent directly via Slack.

vaclav-halik-smartlook (approved):
OK. Nice.

c0mtru1se (approved):
LATM!

PR #268, opened 2025-04-07
DEMRUM-1074: Initialize event

jCharv:
This PR implements the Initialize event as a child span to the AppStart span.


SplunkAgent/Sources/SplunkAgent/Agent/Modules/Manager/AgentModulesManager.swift
```diff
@@ -75,6 +75,21 @@ protocol AgentModulesManager {
     ///
     /// - Parameter metadata: A struct with essential data identification.
     func deleteModuleData(for metadata: any ModuleEventMetadata)
+
+    
+    // MARK: - Metrics
+
+    /// Modules manager needs to provide all modules' initialization timestamps.
```
c0mtru1se, line 82: **suggestion**: I guess this is kinda nit-picky, but should this actually be a job of the module manager? My point is that both the manager and the Agent's install method now have a lot of sorta "floating around" data structures and methods that I think aren't really part of their jobs.

I could imagine some sort of an app / module start times / events "measuring service" or whatever, that would be:
* Accepting the agent's configuration
* Passed / injected to the module manager, so that it can simply just call it's API
* Having all the required data available through its API
* Easily mockable and testable (things like the report structure, etc)
* As a tiny benefit - would be later easily converted to something Swift 6 compatible

Don't have any strong preference here, we can discuss this.

jCharv: @c0mtru1se I thought about some abstraction around these metrics, but I found it to be unnecessary, for now. The metrics in the modules manager are specific for the modules manager, so I decided to keep the implementation simple. However, we could implement something smarter, if we'd need more metrics o more places in the agent and/or in other modules.

As for the Agent's install method, I kept it like that because of simplicity, and because I placed those timestamps arbitrarily, and I count on the fact that those timestamps will very likely change in the future. But I agree that it's not looking pretty atm.

c0mtru1se (approved):

vaclav-halik-smartlook (changes requested):
Overall OK, just need to make some minor adjustments in the code.

vaclav-halik-smartlook (approved):
Perfect!

PR #269, opened 2025-04-07
DEMRUM-1489: Fix session tests

jCharv:
The purpose of this PR was supposed to be fixing the testSessionForLogic() test method. It turned out that the method is not faulty, so the implementation did not change.

This PR also removes the session pulse event.


c0mtru1se (approved):

froseio (approved):
ok

PR #275, opened 2025-04-14
DEMRUM-1199: Add component attribute

jCharv:
This PR implements forcing `component` attribute on all Log events. It also implements attribute checker exporter for all Spans and Logs, which print an error message in case any required attributes are missing.

In this PR the `AgentEvent` base class is removed, and the `Event` protocol is renamed to `AgentEvent`. Therefore all event instances (crash reports event, session start event) now implement the protocol directly instead of subclassing the base class.


vaclav-halik-smartlook (approved):
OK

carlosmcevilly (approved):
lgtm!

I do wonder if this feature could also be a home for length validation for fields (both key length and value length) in user-supplied dictionaries such as those submitted for custom event tracking.

PR #276, opened 2025-04-14
DEMRUM-1201: Add SplunkConfigurationHandler model as a remote configuration bypass

c0mtru1se:
This PR introduces a new configuration class, `SplunkConfigurationHandler`.

It simply serves as a temporary source of truth for our remote config simulation and actual module enablement. The implementation simply hardcodes the configuration so that it can be used and parsed out by the `ModuleManager` without requiring any other changes that could break potentially when O11y cloud implements a proper configuration endpoint.

Please note that this does not break any of our legacy module integration and management as we did not have any local module configuration & config merging APIs in place. Those should be added and resolved later.


froseio (approved):
ok

jCharv (approved):

PR #277, opened 2025-04-14
DEMRUM-930: adding span export to slow rendering module

carlosmcevilly:


SMickelsn (approved):
With the caveat that the approach to settings needs a PM approval, as implemented it seems good to go.

PR #280, opened 2025-04-16
DEMRUM-868 Add server-timing attributs to span

SMickelsn:
Added code to detect "server-timing" headers
Parse header to detect traceId and spanId
Add same to span attributes as link.traceId and link.spanId


carlosmcevilly (approved):
lgtm!

PR #283, opened 2025-04-19
DEMRUM-1683: Implement SessionState and related logic

c0mtru1se:
This PR implements new `SessionState` container as per [these docs](https://splunk.atlassian.net/wiki/spaces/PROD/pages/1078750096708/iOS+agent+API#State) (the links in related ticket are wrong). Note that it does not implement the proposed constructor as it doesn't make contextual sense.

One thing to consider is the unification of docc of the `samplingRate` properties in `SessionState` and `AgentConfiguration`.

### DIFF
--- a/Splunk.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Splunk.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "3eb9999c68975bd1cc59f9b5d725dd6b3d40dee008aa9ee04c4b4fbde03c3c1f",
+  "originHash" : "086cbf8ba05fee02327b46721643451f4c667c178e559c7a6787221339f1d7b4",
   "pins" : [
     {
       "identity" : "grpc-swift",
@@ -31,7 +31,7 @@
     {
       "identity" : "smartlook-ios-sdk-private",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/smartlook/smartlook-ios-sdk-private",
+      "location" : "git@github.com:smartlook/smartlook-ios-sdk-private.git",
       "state" : {
         "branch" : "develop",
         "revision" : "814c1a3e2650ee705ef0cfb0419000397bbf792a"
--- a/SplunkAgent/Sources/SplunkAgent/Public API/API-1.0-Session.swift
+++ b/SplunkAgent/Sources/SplunkAgent/Public API/API-1.0-Session.swift
@@ -27,7 +27,9 @@ public final class Session {
 
     // MARK: - State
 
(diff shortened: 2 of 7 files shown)


SplunkAgent/Sources/SplunkAgent/Public API/API-1.0-Session.swift
```diff
@@ -25,6 +25,10 @@ public final class Session {
     private unowned let owner: SplunkRum
 
 
+    // MARK: - State
+
+    public private(set) lazy var state: SessionState = SessionState(for: owner)
+
```
vaclav-halik-smartlook, line 31: **nitpick:** We always put two empty lines before the second and the following`// MARK:` sections.

SplunkAgent/Sources/SplunkAgent/Public API/API-1.0-Session.swift
```diff
@@ -25,6 +25,10 @@ public final class Session {
     private unowned let owner: SplunkRum
 
 
+    // MARK: - State
+
+    public private(set) lazy var state: SessionState = SessionState(for: owner)
```
vaclav-halik-smartlook, line 30: **chore:** Missing DocC comment.

Splunk.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,5 +1,5 @@
 {
-  "originHash" : "3eb9999c68975bd1cc59f9b5d725dd6b3d40dee008aa9ee04c4b4fbde03c3c1f",
+  "originHash" : "086cbf8ba05fee02327b46721643451f4c667c178e559c7a6787221339f1d7b4",
```
vaclav-halik-smartlook, line 2: **question:** I am not sure about this file (`Package.resolved`). Do we want this file in the repo?

vaclav-halik-smartlook (changes requested):
It looks OK. I have only one question and some minor changes.

Splunk.xcworkspace/xcshareddata/swiftpm/Package.resolved
c0mtru1se: From my pov yes, we do. We've tracked it in CoP also; it makes every dependency update explicit (through the need to commit the update). It can also help CI stability in the future, as we can build with `xcodebuild` and passing the `-disableAutomaticPackageResolution` which will result in the usage of our pins.

What I'm definitely willing to do is to remove this file once we get rid of the `project` and only keep the package's `Package.resolved` file.

vaclav-halik-smartlook: Okay. Deal.

vaclav-halik-smartlook (approved):
OK

PR #285, opened 2025-04-22
Feature/demrum 1996 add mutable attributes

SMickelsn:
- Add MutableAttributes Class
- Add Unit Tests

### DIFF
--- a/SplunkAgent/Sources/SplunkAgent/Public API/Model/MutableAttributes.swift
+++ b/SplunkAgent/Sources/SplunkAgent/Public API/Model/MutableAttributes.swift
@@ -1,0 +1,347 @@
+//
+/*
+Copyright 2025 Splunk Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import Foundation
+import OpenTelemetryApi
+
+class MutableAttributes {
+    private var attributes: [String: AttributeValue]
+
+    // MARK: - Initialize
+
+    init() {
+        attributes = [:]
+    }
+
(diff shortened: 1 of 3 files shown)


SenNeonoveNoci (approved):
From the API’s perspective, it looks good to me.

SplunkAgent/Sources/SplunkAgent/Public API/Model/MutableAttributes.swift
```diff
@@ -0,0 +1,347 @@
...
+import Foundation
+import OpenTelemetryApi
+
+class MutableAttributes {
+    private var attributes: [String: AttributeValue]
+
+    // MARK: - Initialize
+
+    init() {
+        attributes = [:]
```
jzwc, line 27: As Dictionaries in Swift are not thread safe, it might be useful to protect the internal attributes storage with a `DispatchQueue` or similar lock / access serialization mechanism. An example implementation of a possible pattern could be found in [SynchronizedDictionary.swift](https://cd.splunkdev.com/smartlook/smartlook-ios-sdk-private/-/blob/develop/Sources/SmartlookAnalytics/Runtime%20Data/SynchronizedDictionary.swift).

jzwc (changes requested):
Please consider adding some thread saving mechanism.

SplunkAgent/Sources/SplunkAgent/Public API/Model/MutableAttributes.swift
SMickelsn: Thanks Pavel, good suggestion.

jzwc (commented):
Thanks, Stephen. Using a barrier is indeed more nuanced approach than the crude sync used in Smartlook code.  I have no further comments; however, I would appreciate it if a member of the iOS team, who possesses more direct experience with the code, could provide the second approval before merge.

jzwc (approved):

SplunkAgent/Sources/SplunkAgent/Public API/Model/ThreadSafeDictionary.swift
```diff
@@ -0,0 +1,137 @@
...
+    public subscript(key: Key) -> Value? {
+        get {
+            var result: Value?
+            queue.sync {
+                result = dictionary[key]
+            }
+            return result
+        }
+        set {
+            queue.async(flags: .barrier) {
```
carlosmcevilly, line 49: awesome 👍

carlosmcevilly (approved):
looks good! nice tests.

PR #286, opened 2025-04-24
DEMRUM-772 Add IgnoreURLs support to Network Instrumentation

SMickelsn:
- Added IgnoreURLs class to manage Regex patterns supplied by users through the ignoreURLs API (The actual API support to be added in another ticket)
- Added code to stop instrumenting a url if it matches the supplied pattern
- Added unit tests


carlosmcevilly (approved):
looks good!

PR #289, opened 2025-04-26
DEMRUM-1534 Implement standard delegate class limiter

SMickelsn:
Adds an array of standard delegates to pass to URLSessionInstrumentation.

This will avoid the exhaustive search of of all classes in the app that was searching for all possible elements.  This was drastically slowing down the start time of the agent.

For safety, in the case where no items are found, the code will default to the exhaustive search.  This should not happen in practice.


carlosmcevilly (approved):
Looks good!

PR #291, opened 2025-04-28
DEMRUM-1700: Implement api changes

jCharv:
This PR implements DEMRUM-937, DEMRUM-1700 and DEMRUM-1708 by implementing Agent's configuration and api changes and other related changes:

* Rename Agent's singleton to `shared`
* Rename `spanFilter` to `spanInterceptor`
* make spanInterceptor inout to be able to modify spans
* Implement endpoint configuration changes
* Make Agent's `install` method throwing, `install` throws in case of invalid configuration
* Make necessary changes in the agent to support updated configuration
* Update tests


SplunkAgent/Sources/SplunkAgent/Agent/Events/Event Manager/DefaultEventManager.swift
```diff
@@ -86,21 +85,46 @@ class DefaultEventManager: AgentEventManager {
...
 
-        // Initialize log event processor
-        logEventProcessor = OTLPLogEventProcessor(
-            with: configuration.logsUrl,
-            resources: resources,
-            runtimeAttributes: agent.runtimeAttributes,
-            debugEnabled: configuration.enableDebugLogging
-        )
+        // Initialize log event processor. ‼️ Using trace endpoint as a placeholder
+        if let logUrl = configuration.endpoint.traceEndpoint {
```
vaclav-halik-smartlook, line 89: **question:** I don't understand this whole `Noop*` processor logic. There is an expectation that `traceEndpoint` can be `nil`, but it is unclear to me how this could happen since `DefaultEventManager` is never initialized without a valid configuration (see current codebase ). You validate the configuration yourself using the initialization method of the agent instance. 

So, this code is redundant here or am I missing something?

SplunkAgent/Sources/SplunkAgent/Agent/Configuration/Models/AgentConfigurationProtocol.swift
```diff
@@ -35,19 +34,11 @@ protocol AgentConfigurationProtocol: Codable, Equatable {
     var enableDebugLogging: Bool { get set }
     var sessionSamplingRate: Double { get set }
     var globalAttributes: [String: String] { get set }
-    var spanFilter: ((SpanData) -> SpanData?)? { get set }
+    var spanInterceptor: ((inout SpanData) -> SpanData?)? { get set }
```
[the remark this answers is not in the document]
jCharv: @vaclav-halik-smartlook `spanInterceptor` should serve two purposes:
1. Modify the span, this is not possible without the `inout` as `SpanData` has private initializer. It's possible to modify the span with the `SpanData`'s mutating functions only.
2. The output lets the agent know whether to discard the span (by returning `nil`, or whether to send it.

SplunkAgent/Sources/SplunkAgent/Agent/Events/Event Manager/DefaultEventManager.swift
jCharv: @vaclav-halik-smartlook I removed `Noop*` processors, but I needed to handle the `traceUrl` optionality somewhere, so in the update I made `DefaultEventManager`'s init method `throws`. Let me know if this makes sense.

vaclav-halik-smartlook: That's OK now. Thanks!

c0mtru1se (approved):

PR #297, opened 2025-05-07
DEMRUM-938: Implement span interception internals

c0mtru1se:
This PR implements the internals for the already existing public API of the customer-allowed span interception.

It also refactors the `SplunkStdoutSpanExporter` to live in the exporter chain, to correctly log all span data from the whole pipeline.


jCharv (approved):
Nice 👍

PR #298, opened 2025-05-09
DEMRUM-970 Add PLCrashReporter via github

SMickelsn:
- Added PLCrashReporter via github
- Added a separate on device folder to hold the crash reports, to avoid conflict with other instances (Firebase etc.)
- This folder is currently called 'SplunkCrashReports'
- Switched out SPLKPLCrashReporter from the build.

Please note:  Left SplunkCrashReporter in place in the project.  Code is not used and can be removed once testing is complete.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -159,47 +159,11 @@ let package = Package(
         ),
         
         
-        // MARK: SplunkCrashReporter
-        
-        .target(
-            name: "SplunkCrashReporter",
-            path: "SplunkCrashReporter",
-            exclude: [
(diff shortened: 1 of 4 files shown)


jCharv (changes requested):
I think the Agent's `Package.swift` manifest is missing a `PLCrashReporter` dependency. When importing the agent as a swift package, the agent can't be built.

Also when testing the possible code duplication against other PLCrashReporter instances, please use the agent as a swift package dependency in test applications as well, as that will be the way our customers use the agent.

PR #301, opened 2025-05-19
DEMRUM-1401: Fix resources

froseio:
Platform team will finish device model translation on backend. Based on Aditi's comment https://splunk.slack.com/archives/C07PANLTWG1/p1745350583791159

### DIFF
--- a/SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Resources/Resource+AgentResources.swift
+++ b/SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Resources/Resource+AgentResources.swift
@@ -45,7 +45,7 @@ extension Resource {
             ResourceAttributes.deviceModelName.rawValue: .string(agentResources.deviceModelIdentifier),
 
             // OS info
-            ResourceAttributes.osName.rawValue: .string("iOS"),
+            ResourceAttributes.osName.rawValue: .string(agentResources.osName),
             ResourceAttributes.osVersion.rawValue: .string(agentResources.osVersion),
             ResourceAttributes.osDescription.rawValue: .string(agentResources.osDescription),
             ResourceAttributes.osType.rawValue: .string(agentResources.osType)
--- a/SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Resources/Resource+AgentResources.swift
+++ b/SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Resources/Resource+AgentResources.swift
@@ -26,21 +26,8 @@ extension Resource {
     /// Merges Resource object with AgentResources.
     mutating func merge(with agentResources: AgentResources) {
 
-        // Service info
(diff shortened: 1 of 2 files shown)


SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Resources/Resource+AgentResources.swift
```diff
@@ -26,22 +26,9 @@ extension Resource {
     /// Merges Resource object with AgentResources.
     mutating func merge(with agentResources: AgentResources) {
 
-        // Service info
-        // TODO: DEMRUM-1401 - Do we need service name and version?
```
carlosmcevilly, line 30: Maybe you have more up to date info from somewhere, but my interpretation of the top two rows in the first table in the document linked in your ticket (I won't share an internal link here) seems to say that we do need service.name and service.version as they are the OTel semantic convention names that are being used for app (implying app name) and app version (see the middle column, and how it matches up to the left column, not a super perfect match, but it seems to be the intent that we keep these).

jCharv: @carlosmcevilly `service.name` and `service.version` are duplicates of `app` and `app.version`. Backend does not follow semconv, and expects the latter, therefore we don't need to send the former.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Resources/Resource+AgentResources.swift
```diff
@@ -55,12 +42,9 @@ extension Resource {
             ResourceAttributes.deviceManufacturer.rawValue: .string(agentResources.deviceManufacturer),
             ResourceAttributes.deviceId.rawValue: .string(agentResources.deviceID),
 
-            // TODO: DEMRUM-1401 - add translation table from model identifier to model name
             ResourceAttributes.deviceModelName.rawValue: .string(agentResources.deviceModelIdentifier),
 
-
             // OS info
-            // TODO: DEMRUM-1401 - use hardcoded "iOS"?
```
jCharv, line 63: Please perform some testing, whether we need the hardcoded "iOS" value, or whether we can use actual real values like "iPadOS", "tvOS" etc.

I did a quick test, and these additional values seem fine.

jCharv (changes requested):
Please change the PR name to the format of `DEMRUM-xxxx: Name of the PR`, and please check the `os.name` param. Otherwise ok.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Resources/Resource+AgentResources.swift
carlosmcevilly: Thanks for the clarification @jCharv.

jCharv (approved):

carlosmcevilly (approved):
lgtm

PR #306, opened 2025-05-26
DEMRUM-2076: Automatic navigation detection

vaclav-halik-smartlook:
The main logic and surrounding components are designed with the expectation that the `CiscoSwizzling` library will be enhanced to provide improved support for detection. Until those enhancements are implemented, there is a legacy solution in place that relies on capturing system notifications, which is enabled by default. Once the necessary updates to `CiscoSwizzling` are completed, the legacy solution can be safely removed.

**Changes:**
- Automatic navigation detection has been implemented for UIKit-based applications.

**Note:**
- This PR does not include manual detection logic, which will be addressed in a separate PR.
- This PR also does not include unit tests, which will be part of another PR.


jCharv (approved):
@vaclav-halik-smartlook Looks great. One thing I'd suggest is noting the default value in `NavigationModulePreferences.enableAutomatedTracking` docc comment.

PR #307, opened 2025-05-26
DEMRUM-1663: Cleanup

jCharv:
This PR removes some unneeded functionality from the old agent. Also tries to apply linter to all modules, and fixes couple hundred linter issues.

Changes made:

* Fix headers
* Remove unneeded `.swiftpm` and `.gitignore` files
* Remove manual `session.id` parameters
* Remove session start event
* Temporary fix to the "attributeCount" crash
* Update `Package.swift` dependency graph by using OpenTelemetry dependencies directly instead of using `SplunkOpenTelemetry`
* Remove unneeded files
* Enable linter on all modules
* Fix as many linter issues as possible
* Refactor MutableAttributes


vaclav-halik-smartlook (changes requested):
OK. Only a few minor tweaks would be beneficial.

vaclav-halik-smartlook (approved):
OK

PR #310, opened 2025-05-29
remove force try in regex compilation

carlosmcevilly:
Remove try! usage in regex compilation code in addLinkToSpan(::)

- restructure to `guard let regex =`
- introduce log statements for failures: one serious (.fault), one not (.debug)
- add intentional breaking hard failure if regex does not compile. This must be caught early. The objection is we have a hard coded regex; how could it go wrong? And the answer is, of course, editing.
- update the regex to remove the backslashes for double quotes -- although the quotes are nested, this is fine in swift ## regular expression syntax. If I am wrong we will have a preconditionFailure very soon.
- add truncation to the logged string for the .debug log because we get that string from an external source
- add another guard when extracting the traceId and spanId
- move the regex into the function as it's only used in that scope


jCharv:
@carlosmcevilly Could you please update the PR title with the "DEMRUM-xxx" prefix? 🙏

froseio (approved):

carlosmcevilly:
Closing this PR because I've created a replacement. The original work was done before a ticket was created, and the new PR includes commits that mention the ticket number.

PR #312, opened 2025-05-29
DEMRUM-970 Update to PLCrashReporter

SMickelsn:
- Using PLCrashReporter from Github rather than the internal version
- Added a separate on device folder to hold the crash reports, to avoid conflict with other instances (Firebase etc.)
- This folder is currently called 'SplunkCrashReports'
- Switched out SPLKPLCrashReporter from the build.

Please note: Left SplunkCrashReporter in place in the project. Code is not used and can be removed once testing is complete.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,6 +24,10 @@ let package = Package(
         .package(
             url: "https://github.com/open-telemetry/opentelemetry-swift",
             exact: "1.14.0"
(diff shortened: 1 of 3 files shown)


carlosmcevilly (approved):
lgtm

jCharv (approved):
👍

PR #316, opened 2025-06-02
DEMRUM-2090: Implement UserConfiguration and SessionConfiguration

c0mtru1se:
This PR combines the `DEMRUM-2090` and `DEMRUM-2091` tasks into a single API-changing PR.

It also updates all related logic in the Agent, and implements additional test cases for the `user` configuration.


jCharv (approved):
👍

PR #317, opened 2025-06-03
DEMRUM-2077: Custom screen name in Navigation module

vaclav-halik-smartlook:
**Changes:**
- Added logic for custom screen naming.

**Notes:**
- This PR contains all changes from PR #306 and should thus follow after.


SplunkNavigation/Sources/SplunkNavigation/Model/NavigationModel.swift
```diff
@@ -20,19 +20,36 @@ actor NavigationModel {
 
     // MARK: - Public
 
+    public private(set) var moduleEnabled: Bool = true
+
+
```
froseio, line 25: double space

froseio (changes requested):
just a few double spaces and maybe Tasks handeling in property but i need to double check it first

PR #328, opened 2025-06-10
DEMRUM-2370: webview injection fix including legacy support

carlosmcevilly:
implement hybrid sync/async solution for partial legacy support

* the word "partial" is due to legacy BRUM itself; we cannot do everything
  we want with BRUM but we can match the old agent.
* provide legacy synchronous getNativeSessionId() for compatibility.
* add modern async getNativeSessionIdAsync() for improved results
* update DevelApp demo to show both of the above scenarios in action.
* introduce opt-in onNativeSessionIdChanged {} closure property to
  notify BRUM developers when the session ID updates.


SplunkWebView/Sources/SplunkWebView/WebViewInstrumentation.swift
```diff
@@ -0,0 +1,180 @@
...
+
+internal import CiscoLogger
+import SplunkCommon
+import WebKit
+
+public final class WebViewInstrumentation: NSObject {
+
+    private let logger = DefaultLogAgent(poolName: PackageIdentifier.instance(), category: "SplunkWebView")
+
+    public var sharedState: AgentSharedState?
```
c0mtru1se, line 26: **issue**: This property has to be `unowned` to avoid a strong retainment with SplunkRum.

SplunkAgent/Sources/SplunkAgent/Public API/Modules/WebViewInstrumentation/Proxies/Non-Operational/WebViewNonOperational.swift
```diff
@@ -0,0 +1,26 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import WebKit
+
+/// A Non-Operational version to use as a default proxy when the real module isn't installed.
+final class WebViewNonOperational: WebViewInstrumentationModule {
+    func integrateWithBrowserRum(_ view: WKWebView) {
+        // This is a no-op for when the module is not active.
```
c0mtru1se, line 24: **chore**: Please use the `logAccess(toApi: #function)` method here.

Package.swift
```diff
@@ -333,6 +314,7 @@ func generateMainTargets() -> [Target] {
         ),
 
 
+>>>>>>> feature/next-gen
```
c0mtru1se, line 317: **issue**: Please remove this merge resolution leftover, it breaks the compilation.

SplunkWebView/Sources/SplunkWebView/WebViewInstrumentation.swift
```diff
@@ -0,0 +1,180 @@
...
+
+    // NSObject conformance
+    // swiftformat:disable:next modifierOrder
```
c0mtru1se, line 29: **nitpick**: Why is this disable needed?

c0mtru1se (commented):
Couple of small issues & nits, otherwise LGTM. Won't be blocking merging.

SplunkWebView/Sources/SplunkWebView/WebViewInstrumentation.swift
carlosmcevilly: I wonder the same thing. Lint complains about this. Maybe a bug in our linter; I mean to look into that sometime.

SplunkWebView/Sources/SplunkWebView/WebViewInstrumentation.swift
carlosmcevilly: Good catch! Using `weak` since `sharedState` is an optional here.

SplunkAgent/Sources/SplunkAgent/Public API/Modules/WebViewInstrumentation/Proxies/Non-Operational/WebViewNonOperational.swift
carlosmcevilly: added

Package.swift
carlosmcevilly: !! fixed.

SMickelsn (approved):
LGTM.

Nice job on the demo app.

PR #332, opened 2025-06-13
DEMRUM-861: fill out error tracking demo calls

carlosmcevilly:
Improve error tracking api calls in DevelApp

- Add sample attributes to each error
- Add missing stack trace to Error and NSException (for NSError it already worked)
- Add Objective-C code to generate and handle an NSException

Triggering a real NSException gives a more  accurate reflection of real usage where we take content from the callStackSymbols property to populate the stack trace for NSException tracking.

We use Objective-C for generating the exception and this requires that we add bridging header to the DevelApp project.


SMickelsn (approved):
I'm in the middle of a complex change so I didn't run it, but the code looks good to me.

PR #337, opened 2025-06-17
DEMRUM-1756: add legacy report...() wrappers for trackError() and trackEvent() public apis

carlosmcevilly:
Add legacy report...() wrappers for trackError() and trackEvent() public APIs

Public wrappers added:

      reportError(string: String)
      reportError(exception: NSException)
      reportError(error: Error)
      reportEvent(name: String, attributes: NSDictionary)

For reportEvent, added internal convenience NSDictionary init to MutableAttributes in a protocol extension. Also supports Dictionary, for generality, but reportEvent() does not need that.

Added a subset of test buttons in DevelApp to show it's working.


jCharv (changes requested):
Feedback sent in slack.

SMickelsn (approved):
lgtm

PR #341, opened 2025-06-19
DEMRUM-861: cleanups in custom tracking

carlosmcevilly:
* Refactor a property `typeName` to have a better name `eventName` for custom event tracking.
* Fix the handling of `timestamp` in Custom Tracking events and errors.
* Streamline some types, removing unused extensions.
* Combine multiple `track(error:)` calls to remove a redundancy.
* Add code to set OTel attributes for code and namespace if provided in `NSError` `.code` and `.domain`. This can be further improved in the future to also parse the message field but that's not in this code for now.
* Add demo of tracking an `NSError` subclass, whose class name will be reported in attribute `exception.type`.
* Fix lint issues.

This PR is hoped to be the last (or close to the last) changes to the feature branch before we make a PR for merging the `feature/customTracking` branch into `feature/next-gen`. Unit tests are not included yet and will be done separately. The demo app does exercise the features.

If trying this out in Xcode, you'll see stacktrace data in the console; this is intended and is part of the span for errors -- currently not compressed+encoded, but that could be a future step to take.


aditi-s3 (approved):
LGTM

PR #345, opened 2025-06-20
DEMRUM-2436: Add APIs for global attributes

aditi-s3:
- Allow dynamic update of globalAttributes via SplunkRum shared instance (Example added in AgentTestApp ;))
- Allow global Attributes to be passed as a closure for [String: AttributeValue] to keep it simple and pass by reference 
- Simplify processor chaining and use of SplunkStdoutLogExporter in SessionReplay as well as LogToSpan
- Tested e2e and verified global attributes are added to all signals - spans, session replay logs. See screenshots

Screenshot 2025-06-19 at 5 10 29 PM
Screenshot 2025-06-18 at 10 50 51 PM

Note: Unit tests for global Attributes, legacy API mapping not in scope and will be tackled via separate tickets.


SMickelsn (approved):
Tested with the debugger.   Everything seems to work correctly.

PR #349, opened 2025-06-25
DEMRUM-2536: Update log to span exporter's span name resolution logic.

jCharv:
PR expands log to span exporter's span name resolution. No spans should be dropped, therefore we add a default span name, if no span name is provided.

PR also prepares the span name resolution for the `spanName` property, which is not yet implemented in the upstream. This property will take precedence over the `span.name` attribute.


froseio (approved):

vaclav-halik-smartlook (approved):
OK

PR #350, opened 2025-06-25
DEMRUM-2525: Implement support for binary body in Log records.

jCharv:
This PR adds support for Logs binary body.

This is achieved by bypassing the processors-exporters chain in the session replay event processor. A LogRecord with `Data` body support (`SplunkReadableLogRecord`) is created manually and sent to an exporter directly. 

These four files from the upstream are used and modified so that the LogRecord is able to ingest a `Data` body, and that the exporter is able to ingest this modified LogRecord.
- `AttributeValue.swift`
- `CommonAdapter.swift`
- `LogRecordAdapter.swift`
- `ReadableLogRecord.swift`

### DIFF
--- a/SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Log Event Processors/OTLPSessionReplayEventProcessor.swift
+++ b/SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Log Event Processors/OTLPSessionReplayEventProcessor.swift
@@ -15,6 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+internal import CiscoLogger
 import Foundation
 import OpenTelemetryApi
 import OpenTelemetryProtocolExporterCommon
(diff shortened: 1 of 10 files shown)


SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Log Event Processors/OTLPSessionReplayEventProcessor.swift
```diff
@@ -148,18 +150,74 @@ public class OTLPSessionReplayEventProcessor: LogEventProcessor {
...
+
+        // Merge runtime attributes
+        for (key, value) in runtimeAttributes.all {
+            if let attributeValue = SplunkAttributeValue(value) {
+                attributes[key] = attributeValue
+            }
+        }
+
+        // Add attributes from the AgentEvent
+        // ‼️ This code mirrors code from `LogRecordBuilder+AgentEvent.swift`, but utilizes `SplunkAttributeValue`
```
vaclav-halik-smartlook, line 167: **suggestion:** Duplication of code in this form is unfortunate and would require a redesign to some more reasonable form. Attribute handling should be pulled off as a generic function. Then we would have code for all cases.

Please create a new ticket for this change. Thanks!

PR #352, opened 2025-06-27
DEMRUM-1758: add deprecated legacy api mappings for SplunkSlowFrameDetector

carlosmcevilly:
- Add deprecated functions in SplunkRumBuilder for:
  - slow and frozen threshold parameters
  - enabled property
- Update Module code to match.
- Fix module plumbing for isEnabled handling.
- Add unit tests for legacy functions to show module is enabled / disabled as appropriate.

New screenshot from build after latest update (merged latest from feature/next-gen into branch):
Screenshot 2025-07-08 at 12 39 26 AM


SplunkAgent/Sources/SplunkAgent/Public API/Deprecated/Model/API-1.0-SplunkRumBuilder.swift
```diff
@@ -209,16 +247,21 @@ public class SplunkRumBuilder {
             return false
         }
 
-        // Construct module configurations
+        // MARK: - Module Configurations
```
jCharv, line 250: **chore:** Please revert this change. This is a nested MARK, also inside a function. This results in an unnecessary clutter in the symbols menu.

Screenshot 2025-07-07 at 13 12 03

SplunkAgent/Sources/SplunkAgent/Public API/Deprecated/Model/API-1.0-SplunkRumBuilder.swift
```diff
@@ -209,16 +247,21 @@ public class SplunkRumBuilder {
...
+
         var moduleConfigurations: [ModuleConfiguration] = []
 
+        /// Navigation
```
jCharv, line 254: **nitpick:** No need to create docc comments inside an implementation, as this won't generate any symbol documentation. Just use `//`.

SplunkAgent/Sources/SplunkAgent/Public API/Deprecated/Model/API-1.0-SplunkRumBuilder.swift
```diff
@@ -209,16 +247,21 @@ public class SplunkRumBuilder {
...
         let navigationModuleConfiguration = SplunkNavigation.NavigationConfiguration(
             isEnabled: screenNameSpans,
             enableAutomatedTracking: showVCInstrumentation
         )
-
         moduleConfigurations.append(navigationModuleConfiguration)
 
+        /// SlowFrameDetector
```
jCharv, line 261: **nitpick:** Same as above, no need to create docc comments inside an implementation, as this won't generate any symbol documentation. Just use `//`.

jCharv (changes requested):
Just couple of small nits, which I think should be addressed. Otherwise looking good.

jCharv (approved):

PR #360, opened 2025-07-04
NO-TICKET: Fix resource object assignment in release builds

jCharv:
This PR fixes a bug where in the `OTLPSessionReplayEventProcessor` we are storing the `resource` object for resources to be added into all session replay logs manually. This assignment failed in release builds as the resource object assignment was done inside a debug macro.


carlosmcevilly (approved):
lgtm

PR #364, opened 2025-07-09
DEMRUM-2684: Fix package build warnings.

jCharv:
This PR fixes couple of warnings when building the agent as a swift package (when opening the package in Xcode directly, instead of using it as a dependency in a test application).

This PR does not fix linter issues or other warnings when the package is opened from the TestApp or the DevelApp.


vaclav-halik-smartlook (approved):
OK

PR #376, opened 2025-07-29
DEMRUM-2428: Update Network Span Attributes

SMickelsn:
Update Network Attributes to new semantics

- update content_length attributes
- update url attributes
- update scheme attribute
- add support to remove obsolete keys from OTel instrumentation
- update unit tests

### DIFF
--- a/SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
+++ b/SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
@@ -164,9 +164,15 @@ public class NetworkInstrumentation {
         let body = URLRequest.httpBody
         let length = body?.count ?? 0
         span.setAttribute(key: key, value: length)
+        let method = URLRequest.httpMethod ?? "unknown_method"
+        span.setAttribute(key: SemanticAttributes.httpRequestMethod, value: method)
         span.setAttribute(key: "component", value: "http")
 
         if let url = URLRequest.url {
+            if let scheme = url.scheme {
+                span.setAttribute(key: SemanticAttributes.urlScheme, value: scheme)
+            }
+
             if let host = url.host {
                 span.setAttribute(key: "server.address", value: host)
                 // Preload with host in case IP cannot be determined
@@ -264,6 +270,9 @@ public class NetworkInstrumentation {
             }
         }
 
+        // removes obsolete attributes
+        removeObsoleteAttributesFromSpan(span)
+
         /* Save this until we add the feature into the Agent side API
         guard ((agentConfiguration?.appDCloudNetworkResponseCallback?(URLResponse)) == nil) else {
             let newUrl = ((agentConfiguration?.appDCloudNetworkResponseCallback!(URLResponse)) != nil)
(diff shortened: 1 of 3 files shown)


SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
```diff
@@ -327,6 +338,25 @@ public class NetworkInstrumentation {
         return false
     }
 
+    /// Removes obsolete attributes from the span
+    /// - Parameter span: The span to be modified
+    private func removeObsoleteAttributesFromSpan(_ span: Span) {
+        // Attributes to be removed
+        let attributesToRemove = [
+            "http.url",
```
aditi-s3, line 346: Instead of hard coding attribute keys directly, we should use the SemanticAttributes enum from upstream to define which attributes to remove from attribute map. For example use SemanticAttributes.httpMethod instead of "http.method"

SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
```diff
@@ -160,13 +160,21 @@ public class NetworkInstrumentation {
...
+        let key = SemanticAttributes.httpRequestBodySize
         let body = URLRequest.httpBody
         let length = body?.count ?? 0
         span.setAttribute(key: key, value: length)
+        let method = URLRequest.httpMethod ?? "unknown_method"
+        span.setAttribute(key: SemanticAttributes.httpRequestMethod, value: method)
         span.setAttribute(key: "component", value: "http")
 
         if let url = URLRequest.url {
+            span.setAttribute(key: SemanticAttributes.urlPath, value: url.path)
```
carlosmcevilly, line 172: `URL` `.path`  is deprecated in favor of [path(percentencoded:)](https://developer.apple.com/documentation/foundation/url/path(percentencoded:))

SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
```diff
@@ -160,13 +160,21 @@ public class NetworkInstrumentation {
...
+            span.setAttribute(key: SemanticAttributes.urlQuery, value: url.query ?? "")
```
carlosmcevilly, line 173: `.query` also has that same [deprecation treatment](https://developer.apple.com/documentation/foundation/url) also `.host`, although that one is not part of your PR.

SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
```diff
@@ -327,6 +338,25 @@ public class NetworkInstrumentation {
         return false
     }
 
+    /// Removes obsolete attributes from the span
+    /// - Parameter span: The span to be modified
+    private func removeObsoleteAttributesFromSpan(_ span: Span) {
```
carlosmcevilly, line 343: nit: internal function so customers will not see it, but a more swifty option would be

    `private func removeObsoleteAttributes(from span: Span) {`

Then of course the call site would become `removeObsoleteAttributes(from: span)`

SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
```diff
@@ -264,6 +272,9 @@ public class NetworkInstrumentation {
             }
         }
 
+        // removes obsolete attributes
+        removeObsoleteAttributesFromSpan(span)
```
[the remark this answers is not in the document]
SMickelsn: Good catch.   Added to error path

carlosmcevilly (approved):

SplunkNetwork/Sources/SplunkNetwork/NetworkInstrumentation.swift
```diff
@@ -327,12 +338,34 @@ public class NetworkInstrumentation {
         return false
     }
 
+    /// Removes obsolete attributes from the span
```
aditi-s3, line 341: non-public method - we should use `//` instead

PR #378, opened 2025-07-30
DEMRUM-2771: Fix session rotation

jCharv:
Session rotation was not happening immediately after receiving the `willEnterForeground` notification. This caused some data (e.g. app start, session replay)  using session id from a previous session. This PR fixes this issue.

The PR also fixes an issue in the default session tests.

### DIFF
--- a/SplunkAgent/Sources/SplunkAgent/Agent/Session/DefaultSession.swift
+++ b/SplunkAgent/Sources/SplunkAgent/Agent/Session/DefaultSession.swift
@@ -334,6 +334,9 @@ extension DefaultSession {
                 // We need to mark the time of this transition
                 if self?.enterBackground != nil {
                     self?.leaveBackground = Date()
+
+                    // Refresh session immediately
+                    self?.refreshSession()
                 }
             }
 
--- a/SplunkAgent/Tests/SplunkAgentTests/Agent/Session/DefaultSessionTests.swift
+++ b/SplunkAgent/Tests/SplunkAgentTests/Agent/Session/DefaultSessionTests.swift
@@ -242,9 +242,8 @@ extension DefaultSessionTests {
 
         /* Going into the background for *too long* time */
         try simulateBackgroundStay(for: defaultSession, duration: 12)
-        simulateMainThreadWait(duration: 2)
-
-        // The current session should *not* be the same
+ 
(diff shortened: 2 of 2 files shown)


SplunkAgent/Sources/SplunkAgent/Agent/Session/DefaultSession.swift
```diff
@@ -332,6 +334,9 @@ extension DefaultSession {
                 // We need to mark the time of this transition
                 if self?.enterBackground != nil {
                     self?.leaveBackground = Date()
+
+                    // Refresh session immediately
+                    self?.refreshSession()
```
carlosmcevilly, line 339: Approve... with other concerns.

The small change makes sense.

Looking at the whole file, there are some issues,  but not part of your change and they can probably be separated out as their own ticket.

I think accessQueue is severely underused in here. It should be used any time we are changing the session, and also when changing things upon which we decide whether to change the session like enterBackground and leaveBackground values. Calls to refreshSession() could also be dispatched to that queue and maybe some other things.

One also wonders whether we should reset the enterBackground / leaveBackground upon checking them, but the benefits would be pretty small. That's what got me thinking about concurrency, because of a race condition that would unfortunately be made worse by such an "improvement."

carlosmcevilly (approved):

SplunkAgent/Sources/SplunkAgent/Agent/Session/DefaultSession.swift
vaclav-halik-smartlook: @carlosmcevilly I don't understand your comments; I don't know where the weak points you mention are. Perhaps it would be worth explaining in more detail.

The current implementation does not need an additional layer of protection, as the data is only changed by the class that owns it. The data is of a value type, which also helps in this case (but not as significantly as it might seem). There is a fully functional barrier using synchronization on the read side. This synchronization ensures that there will never be any unwanted behavior or problems when accessing from multiple threads.

You can try it yourself by reducing the interval for `maxSessionLength` and, for example, running parallel queries on the session ID. I have now tried running more than 10,000,000 queries, and everything behaves as expected.

vaclav-halik-smartlook (approved):
Nice fix!

PR #382, opened 2025-07-31
DEMRUM-2528: Add Screen Name to Crash Reports

SMickelsn:
Inital call from Navigation to Crash Reports

### DIFF
--- a/SplunkCrashReports/Sources/SplunkCrashReports/CrashReports.swift
+++ b/SplunkCrashReports/Sources/SplunkCrashReports/CrashReports.swift
@@ -213,6 +213,11 @@ public class CrashReports {
         }
     }
 
+    public func crashReportUpdateScreenName(_ screenName: Any) {
+        deviceDataDictionary["screenName"] = "\(screenName)"
+        updateDeviceStats()
+    }
+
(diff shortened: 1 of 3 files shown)


SplunkCrashReports/Sources/SplunkCrashReports/CrashReports.swift
```diff
@@ -283,6 +288,7 @@ public class CrashReports {
                 reportDict[.batteryLevel] = unarchivedData["battery"]
                 reportDict[.freeMemory] = unarchivedData["disk"]
                 reportDict[.freeDiskSpace] = unarchivedData["memory"]
+                reportDict[.screenName] = unarchivedData["screenName"]
```
carlosmcevilly, line 291: Should this also be gated like this?

`#if canImport(SplunkCrashReports)`

The only place I see where that key gets set is inside an if like that, so I wonder if the access is problematic without the if.

SMickelsn: Don't need that here as this code is actually in SplunkCrashReports.   The ifdef is only in the SplunkAgent because the Crash Module might not be linked in.

carlosmcevilly (approved):

PR #390, opened 2025-08-11
DEMRUM-2905: Objective-C API for Global Attributes

vaclav-halik-smartlook:
**Changes:** 
- Implemented a basic version of `AttributeValue` in Objective-C.
- Included missing logic for handling global attributes in Objective-C projects.
- Improved accessibility of error messages related to agent configuration.

### DIFF
--- a/SplunkAgent/Sources/SplunkAgentObjC/Model/API-1.0-AttributeValueObjC.swift
+++ b/SplunkAgent/Sources/SplunkAgentObjC/Model/API-1.0-AttributeValueObjC.swift
@@ -39,9 +39,10 @@ public final class AttributeValueObjC: NSObject {
 
     /// Initializes a new instance of an attribute for a `BOOL` type value.
     ///
-    /// - Parameter bool: The value for the new attribute.
-    public convenience init(bool: Bool) {
-        self.init(boolNumber: NSNumber(value: bool))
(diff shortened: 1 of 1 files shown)


froseio (approved):

SplunkAgent/Sources/SplunkAgentObjC/Model/API-1.0-AttributeValueObjC.swift
```diff
@@ -0,0 +1,269 @@
...
+
+    /// Identifies the type of attribute value held.
+    public let type: AttributeValueTypeObjC
+
+
+    // MARK: - Initialization
+
+    /// Initializes a new instance of an attribute for a `BOOL` type value.
+    ///
+    /// - Parameter bool: The value for the new attribute.
```
jCharv, line 42: **chore:** The `bool` param name in this method gets replaced by the compiler to `bool_` in the generated header file, resulting in a warning in a test app. This is also a case for the `double` variants. 

I guess we can either suffix the docc param with `_`, or we can rename these params altogether. Up to you.
Screenshot 2025-08-14 at 9 53 28

jCharv (changes requested):
Looks great, we should just fix the parameter names to fix warnings in the generated -Swift.h file.

jCharv (approved):

PR #392, opened 2025-08-11
DEMRUM-2683: Update SetAttributes Usage

SMickelsn:
- Update Network Module
- Update AppStart Module
- Update Crash Report Module
- Update Navigation Module
- Update Network Monitor Module
- Update Span Processors
- Update Slow Frame Detector

Each of the above modules was modified as requested to clear attributes before adding them.  In some cases, it was not entirely clear that the treatment is required but given safety of doing it and recommendations it seemed ok to err on that side.  I can certainly undo any module's changes.

The change has been implemented in a span extension as recommended by Vaclav


vaclav-halik-smartlook:
Hi @SMickelsn , I think it looks great. I like the use of `@_spi` for this extension. Nice work and thanks for the changes!

PR #394, opened 2025-08-14
DEMRUM-781: Fix/lint formatting issues

aditi-s3:


Applications/AgentTestApp/AgentTestApp/Features/NetworkCalls.swift
```diff
@@ -18,16 +18,15 @@ limitations under the License.
...
-        
+
         guard let url = URL(string: targetURL) else { return }
         let request = URLRequest(url: url)
         let semaphore = DispatchSemaphore(value: 0)
 
         let task = URLSession.shared.dataTask(with: request) { data, _, _ in
-            if let data = data {
-                let string = String(decoding: data, as: UTF8.self)
+            if let data = data, let string = String(data: data, encoding: .utf8) {
```
vaclav-halik-smartlook, line 29: **chore:** The correct formatting should be:
```swift
if 
    let data = data,
    let string = String(data: data, encoding: .utf8) 
{
```

Applications/AgentTestApp/AgentTestApp/Features/NetworkCalls.swift
```diff
@@ -18,16 +18,15 @@ limitations under the License.
 import Foundation
 
 class NetworkCalls {
-    
+
     func simpleNetworkCallWith(targetURL: String) {
-        
+
         guard let url = URL(string: targetURL) else { return }
```
vaclav-halik-smartlook, line 24: chore: The correct formatting should be:
```swift
guard let url = URL(string: targetURL) else { 
    return
}
```

SplunkAgent/Sources/SplunkAgent/Utils/Repeating jobs/RepeatingJob.swift
```diff
@@ -83,6 +83,7 @@ class RepeatingJob: AgentRepeatingJob {
     // MARK: - Private methods
 
     private func createTimer() -> Timer {
+        // swiftlint:disable:next unowned_variable_capture
```
vaclav-halik-smartlook, line 86: **issue:** Disable for SwiftLint rule *must* always be in pair with the following enable.

vaclav-halik-smartlook (changes requested):
Not OK. Multiple problems to solve.

SplunkAgent/Sources/SplunkAgent/Utils/Repeating jobs/RepeatingJob.swift
vaclav-halik-smartlook: That makes sense. Good point.

PR #420, opened 2025-08-26
DEMRUM-3041: Integrate latest Session Replay binaries

mbudinsky:
This PR updates the prebuilt Session Replay binaries to build `1.0.6.207`, and updates our internal navigation event models accordingly.


vaclav-halik-smartlook (approved):
OK

PR #634, opened 2026-05-04
DEMRUM-5470: Use event timestamp for presentation transitions

carlosmcevilly:
### Description

* `PresentationTransitionSnapshot` was creating `NavigationPair` and synthetic end-event timestamps via `Date()`, capturing async-handler execution time rather than when UIKit signaled the transition.
* Added `timestamp: Date` to `PresentationTransitionSnapshot`; replaced all four `Date()` call sites with `snapshot.timestamp`.
* Added `testPresentationTransitionStartPreservesEventTimestamp` to verify the timestamp is forwarded correctly.

### Checklist

- [ ] I have added an entry to CHANGELOG for any user facing changes.
- [ ] \(Optional) I have updated any relevant documentation.

### Generative AI usage

### How to Test These Changes

Includes unit tests.


jCharv (approved):

jCharv (changes requested):
**question:** I noticed that `processNavigationEnd()` ignores `event.timestamp` and sets the `let end = Date()`. Is this correct behaviour?

carlosmcevilly:
> **question:** I noticed that `processNavigationEnd()` ignores `event.timestamp` and sets the `let end = Date()`. Is this correct behaviour?

Yes, intentional. These events fire upon transition completion, so Date() is the right timestamp for the end.

SplunkNavigation/Sources/SplunkNavigation/Navigation+PresentationControllerTransitions.swift
```diff
@@ -103,7 +106,7 @@ extension Navigation {
 
         let navigation = NavigationPair(
             type: .transition,
-            start: Date(),
+            start: snapshot.timestamp,
```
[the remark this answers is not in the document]
carlosmcevilly: No CHANGELOG entry needed here. This corrects an implementation bug in unreleased code.

jCharv:
@carlosmcevilly I still don't understand. The span start timestamp is set at the navigation end, right? The span end date is set at time when the `processNavigationEnd` happens to run. So the span measures the duration of the async execution time. I don't see why this information should be useful to the customer? Am I missing something here?

carlosmcevilly:
@jCharv I looked into this more; you were right, good catch. Updated. The span end now uses the timestamp captured in the swizzled presentationDidEnd, so the duration more closely reflects actual animation time rather than collapsing to near zero.

SplunkNavigation/Sources/SplunkNavigation/Navigation.swift
```diff
@@ -292,7 +292,7 @@ public final class Navigation: Sendable {
 
     /// Process the finalizing of the navigation.
     func processNavigationEnd(event: NavigationActionEvent) async {
-        let end = Date()
+        let end = event.timestamp
```
jCharv, line 250: **issue:** Now the issue is that the `end` timestamp can be set to an earlier date than the start timestamp in `updateTransitionStart()`, if the async `updateTransitionStart()` method is run later than expected. The result would be a negative duration. I'd consider using the `event.timestamp` for all these `start: Date()` occurrences as well.

jCharv (changes requested):

SplunkNavigation/Sources/SplunkNavigation/Navigation.swift
carlosmcevilly: Good point; fixed.

SMickelsn (approved):

SplunkNavigation/Sources/SplunkNavigation/Navigation.swift
jCharv: There are still some remaining `start: Date()` which haven't been fixed.

carlosmcevilly: Thanks, I think they are fixed now.

PR #646, opened 2026-05-07
DEMRUM-5269: Add userActivity timestamps to session replay metadata for timeline visualization

frose-dev:
### Description

Adds a `userActivity` field to the `segmentMetadata` JSON payload of Session Replay data events. The field contains Unix-millisecond timestamps of every user interaction captured by the `Interactions` module within the segment window, so the backend can render a timeline of user activity alongside the replay.

Design summary:

- `Interactions` exposes a synchronous `onActivity` callback (SPI-scoped), invoked directly from `handleEvent` before type-specific processing. The property is guarded by an `NSLock` so the host-app install path and the detector task can safely read/write it concurrently.
- `UserActivityCollector` accumulates timestamps as Unix milliseconds. It is lock-protected and gated by an explicit recording flag so events never accumulate while Session Replay is not recording (sampled out, disabled, or paused via `stop()`). The buffer is capped at 10 000 with an amortized trim slack.
- `flush(startMs:endMs:)` returns timestamps `<= endMs` alongside a generation token. Late-arriving events whose real timestamp falls in an already-flushed window ride the next segment rather than being orphaned.
- `restore(_:generation:)` re-inserts timestamps when a send fails, but is a no-op if a stop/reset changed the generation in between — a delayed failure callback cannot smuggle the previous recording session's activity into a new one.
- Lifecycle is driven by `SessionReplay.start()` / `stop()`:
  - `start()` calls `reset()` (clears state, bumps generation) then enables recording.
  - `stop()` disables recording synchronously; `module.stop()` then closes the final record and its activity is still flushed by `publishSessionReplay` (no eager `reset()` on stop).
  - At install time the collector is primed to accept events because the Cisco SessionReplay module auto-starts recording as part of install.
- `publishSessionReplay` in `DefaultEventManager` enriches the encoded metadata with `userActivity` and restores the flushed timestamps on the retry path with the returned generation token.

### Behavior notes

- When Session Replay is disabled by configuration, sampled out, or the Session Replay module is not installed, no wiring is set up and no timestamps are collected.
- When `InteractionsConfiguration.isEnabled` is `false`, the detector never runs, so no timestamps are emitted — the `userActivity` array on each segment is empty. This is documented in the changelog; source of truth for interaction detection remains the Interactions module.

### Checklist

- [x] (Optional) I have added unit tests for my changes.
- [ ] (Optional) I have updated the sample app for integration testing.
- [ ] (Optional) I have updated any relevant documentation.

### Generative AI usage

### How to Test These Changes

Setup:

1. Open a test app that uses the SDK (e.g. `SplunkRumWorkspace/TestApp`) in Xcode.
2. Configure the agent with a valid endpoint or point it at a local proxy.
3. Enable Session Replay in the agent configuration.
4. Run the app on a simulator or device.

Steps:

1. Launch the app and wait for session replay recording to start.
2. Perform several taps and other gestures on the UI.
3. Wait for a Session Replay segment to be exported (segments are flushed periodically — you can trigger one sooner by backgrounding the app or by calling `startNewDataChunk()`).
4. Inspect the outgoing OTLP log export request (via proxy or backend).

What to verify in the payload:

Locate the log record for `session_replay_data`. Find the `segmentMetadata` attribute and confirm it contains a `userActivity` array with Unix millisecond timestamps corresponding to your taps:

```json
{
  "startUnixMs":  1700000000000,
  "endUnixMs":    1700000005000,
  "source":       "...",
  "userActivity": [1700000001123, 1700000002456, 1700000003789]
}
```

- The array should be empty (`[]`) if no interactions were performed between segment flushes.
- Performing interactions in one segment and verifying that the next segment has an empty array confirms the buffer is correctly flushed per segment.

Lifecycle checks:

- Call `SplunkRum.shared.sessionReplay.stop()`; further interactions must not be attached to any segment.
- Call `SplunkRum.shared.sessionReplay.start()` again; interactions from the previous stopped period must not reappear on new segments.

Backwards compatibility:

Old payloads without `userActivity` are still decoded correctly by existing decoders — `Metadata` decoding tolerates the missing field in both directions.


SplunkAgent/Sources/SplunkAgent/Agent/Events/Event Manager/DefaultEventManager.swift
```diff
@@ -196,29 +198,23 @@ class DefaultEventManager: AgentEventManager {
             // Use scriptInstanceId as a 16 character substring of a sessionId
             let scriptInstanceId = String(sessionId.prefix(upTo: sessionId.index(sessionId.startIndex, offsetBy: 16)))
 
+            let userActivity = await userActivityCollector.flush()
```
jCharv: **issue:** Any userActivity events happening between session replay's end and this call are added to this session replay record due to it being asynchronous. Imo you should use `metadata.startUnixMs` and `metadata.endUnixMs` to determine which userActivity to include in the record.

SplunkAgent/Sources/SplunkAgent/Agent/Modules/SplunkRum+Modules.swift
```diff
@@ -272,6 +272,15 @@ extension SplunkRum {
             return
         }
 
+        // Forward interaction timestamps to the event manager's collector
+        if let collector = (eventManager as? DefaultEventManager)?.userActivityCollector {
```
jCharv: **issue:** Collector collects unbounded even when session replay is disabled and/or sampled out.

SplunkAgent/Sources/SplunkAgent/Public API/SplunkRum.swift
jCharv, line 1: **question:** Are changes in this PR user facing in any way? Would it make sense to put an item into the CHANGELOG?

jCharv (changes requested):

SplunkAgent/Sources/SplunkAgent/Agent/Events/Event Manager/DefaultEventManager.swift
```diff
@@ -196,29 +198,23 @@ class DefaultEventManager: AgentEventManager {
             // Use scriptInstanceId as a 16 character substring of a sessionId
             let scriptInstanceId = String(sessionId.prefix(upTo: sessionId.index(sessionId.startIndex, offsetBy: 16)))
 
+            let userActivity = await userActivityCollector.flush(startMs: metadata.startUnixMs, endMs: metadata.endUnixMs)
```
vaclav-halik-smartlook: [P2] These segment callbacks are handled by independent `Task`s, so their execution order is not guaranteed. If segment B flushes before earlier segment A, `flush` removes every timestamp `<= B.endMs`, including A's range; additionally, timestamps are removed before `sendEvent` reports success, so a failed or recovered/replayed chunk cannot retain them. Please serialize segment assignment or persist/key activity per segment and discard it only once the corresponding chunk has been accepted.

SplunkAgent/Sources/SplunkAgent/Agent/Modules/SplunkRum+Modules.swift
```diff
@@ -272,6 +272,16 @@ extension SplunkRum {
             return
         }
 
+        if let collector = (eventManager as? DefaultEventManager)?.userActivityCollector,
+            sessionReplayProxy is SessionReplay
+        {
+            userActivityTask = Task {
+                for await date in interactionsModule.activityStream {
+                    await collector.record(at: date)
```
vaclav-halik-smartlook: [P2] `yield` and this `record` happen on different tasks. A timestamp yielded before a segment closes can reach the collector only after that segment has flushed, making it too old for the next segment and silently losing it. Please place recording and segment flushing on one ordered execution path, or establish an equivalent happens-before guarantee.

SplunkAgent/Sources/SplunkAgent/Agent/Events/Utils/UserActivityCollector.swift
```diff
@@ -0,0 +1,46 @@
...
+import Foundation
+
+/// Collects user interaction timestamps for inclusion in session replay metadata.
+///
+/// Timestamps are stored as Unix milliseconds and flushed when a session replay segment is produced.
+actor UserActivityCollector {
+
+    // MARK: - Private properties
+
+    private var timestamps: [Int] = []
```
vaclav-halik-smartlook, line 37: [P2] This buffer remains unbounded whenever the real Session Replay proxy exists but recording has not started, is paused, or has stopped; the consumer continues to append while no segment flush occurs. Please tie collection to the recording lifecycle and add a bounded retention/reset policy.

SplunkInteractions/Sources/SplunkInteractions/Module/Interactions.swift
```diff
@@ -42,19 +55,16 @@ public final class Interactions: SplunkInteractionsModule {
...
-    // MARK: - Internal properties
-
-    var interactionsDetector: InteractionsDetector<DefaultSwizzling>?
-
-
     // MARK: - Initialization
 
     public required init() {
         destination = OTelDestination()
+        (activityStream, activityContinuation) = AsyncStream.makeStream(bufferingPolicy: .bufferingNewest(1_000))
```
vaclav-halik-smartlook: [P3] `.bufferingNewest(1_000)` can drop older interactions when the consumer falls behind, so the implementation is not lossless despite the documentation and PR description claiming every event. Please either provide a lossless ordered path or explicitly define best-effort semantics and test overflow behavior.

SplunkAgent/Tests/SplunkAgentTests/Agent/Events/EventsTests.swift
```diff
@@ -74,6 +75,47 @@ final class EventsTests: XCTestCase {
         wait(for: [requestExpectation], timeout: 10)
     }
 
+    func testSessionReplayDataEventUserActivityInMetadata() throws {
```
vaclav-halik-smartlook, line 78: [P3] These tests inject an already-prepared `[Int]` directly into `SessionReplayDataEvent`, so they only verify JSON serialization and do not exercise the new pipeline. Please add coverage through the activity stream, collector, and segment publication for window partitioning, out-of-order segments, delayed consumption, inactive recording, and failed or recovered chunks.

SplunkAgent/Sources/SplunkAgent/Agent/Events/Events/SessionReplayDataEvent.swift
```diff
@@ -80,14 +81,41 @@ class SessionReplayDataEvent: AgentEvent {
...
+            "endUnixMs": metadata.endUnixMs,
+            "source": metadata.source,
+            "userActivity": userActivity
+        ]
+        guard
+            let data = try? JSONSerialization.data(withJSONObject: dict),
+            let jsonString = String(data: data, encoding: .utf8)
+        else {
+            let activityJSON = userActivity.map(String.init).joined(separator: ",")
+            return "{\"startUnixMs\":\(metadata.startUnixMs),\"endUnixMs\":\(metadata.endUnixMs),\"userActivity\":[\(activityJSON)]}"
```
vaclav-halik-smartlook: [P4] The last-resort fallback still omits `source`, changing the metadata schema if this path runs. Please include and correctly escape `source`, or replace the manual string construction with a fallback that preserves the same fields as the normal path.

SplunkNavigation/Tests/SplunkNavigationTests/NavigationControllerNameTests.swift
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 
 import XCTest
 
+// swift-format-ignore-file
```
vaclav-halik-smartlook: [P4] `// swift-format-ignore-file` must appear before any import or code; in this position it is ignored. Please move it to the top of the file or remove it, since the file currently passes the formatter without it.

SplunkCommon/Tests/SplunkCommonTests/Persistent Cache/PersistentCacheModelTests.swift
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 
 import XCTest
 
+// swift-format-ignore-file
```
vaclav-halik-smartlook: [P4] `// swift-format-ignore-file` must appear before any import or code; in this position it is ignored. Please move it to the top of the file or remove it, since the file currently passes the formatter without it.

vaclav-halik-smartlook (changes requested):
Please address the ordering, lifecycle, retry, and test-coverage issues called out inline before merging. The current implementation can silently lose or retain user-activity timestamps even though the existing test suite passes.

vaclav-halik-smartlook (approved):
OK

PR #683, opened 2026-06-25
DEMRUM-5839: Add Request Marker for Internal Network Requests

SMickelsn:
### Description

Internal SDK calls to the RUM platform are being instrumented and emitted as network telemetry. This suggests the network exclusion logic may not be consistently suppressing internal SDK/RUM platform requests.

- Add an internal-request marker in the exporter request creation path and have SplunkNetwork skip those before URL matching.
- Retain current endpoint exclusion logic to support IngoreURLs

### Checklist

- [ ] I have added an entry to CHANGELOG for any user facing changes.
- [ ] \(Optional) I have added unit tests for my changes.
- [ ] \(Optional) I have updated any relevant documentation.

### Generative AI usage

### How to Test These Changes

[Provide clear and step-by-step instructions on how reviewers can test the changes you've made.]

[Mention any related upcoming work or considerations for the future.]

### DIFF
--- a/SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+Span.swift
+++ b/SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+Span.swift
@@ -36,6 +36,10 @@ func startHttpSpan(request: URLRequest?) -> Span? {
         return nil
     }
 
+    if TraceContextInjector.hasTraceContext(in: request) {
+        return nil
+    }
+
     if !(url.scheme?.lowercased().starts(with: "http") ?? false) {
         return nil
     }
--- a/SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+Span.swift
+++ b/SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+Span.swift
@@ -29,6 +29,13 @@ func startHttpSpan(request: URLRequest?) -> Span? {
         return nil
     }
 
+    if InternalNetworkRequestMarker.isMarked(request) {
+        NetworkInstrumentationManager.shared.logger.log(level: .debug) {
+            "URL excluded as an internal SDK request \(url.absoluteString)"
+        }
+        return nil
+    }
+
     if !(url.scheme?.lowercased().starts(with: "http") ?? false) {
         return nil
     }
--- a/SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+Span.swift
+++ b/SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+Span.swift
@@ -36,10 +36,6 @@ func startHttpSpan(request: URLRequest?) -> Span? {
         return nil
     }
 
-    if TraceContextInjector.hasTraceContext(in: request) {
-        return nil
-    }
-
     if !(url.scheme?.lowercased().starts(with: "http") ?? false) {
         return nil
     }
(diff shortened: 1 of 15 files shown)


SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+Span.swift
```diff
@@ -29,6 +29,13 @@ func startHttpSpan(request: URLRequest?) -> Span? {
         return nil
     }
 
+    if InternalNetworkRequestMarker.isMarked(request) {
```
aditi-s3, line 32: @SMickelsn  Is this marker check needed here if we have the check in startHttp span https://github.com/signalfx/splunk-otel-ios/pull/683/changes#diff-1c0d9d854f8f9d7ff8472ba2de69c77d425cd91c0844ee8579e351ae33f19608R32

SMickelsn: I think this guard should stay in startHttpSpan because that function is also used by the resume-time fallback for tasks that were not instrumented at creation time. Those calls do not pass through shouldInstrumentRequest, so startHttpSpan needs to remain the final centralized exclusion gate for SDK-owned requests.

SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+TaskCreation.swift
```diff
@@ -193,26 +195,66 @@ func createInstrumentedDownloadTask(
...
+}
+
+func instrumentationDecision(for request: URLRequest) -> RequestInstrumentationDecision {
+    if InternalNetworkRequestMarker.isMarked(request) {
+        return .skipInstrumentation
     }
 
-    guard let span = startHttpSpan(request: request) else {
-        return originalIMP(session, selector, request)
+    if TraceContextInjector.hasTraceContext(in: request) {
```
carlosmcevilly, line 215: (With AI assistance here) It still seems like `traceparent` may need its own case instead of being folded into `.skipInstrumentation` unless you really mean to skip every request that already includes a `traceparent` (even those from possible non-Splunk injection of `traceparent` items).

The internal marker already added in this PR is a strong local signal that could be used instead, because the exporter sets it itself on SDK-owned requests:

  ```swift
  URLProtocol.setProperty(
      true,
      forKey: "com.splunk.rum.internal-network-request",
      in: mutableRequest
  )
```

So this case makes sense:

  ```swift
  if InternalNetworkRequestMarker.isMarked(request) {
      return .skipInstrumentation
  }
```

But a pre-existing `traceparent` only means "this request already has trace context.' It does not necessarily mean this request is SDK-owned, or that Splunk should drop customer network telemetry for it.

Could we add a third decision?

  ```swift
  enum RequestInstrumentationDecision {
      case instrumentAtCreation
      case deferToResume
      case skipInstrumentation
  }
```

Then classify requests like this:

  ```swift
  if InternalNetworkRequestMarker.isMarked(request) {
      return .skipInstrumentation
  }

  if TraceContextInjector.hasTraceContext(in: request) {
      return .deferToResume
  }

  return .instrumentAtCreation
```

(or a `switch` if you prefer but an `if` chain is pretty readable here).

For `.deferToResume`, task creation would just create the original task without injecting headers and without marking it skipped:

  ```swift
  case .deferToResume:
      return createOriginalTask()
```

That preserves the existing `traceparent`, but still lets the existing `resume()` fallback create a Splunk HTTP span from `currentRequest`.

Related point: for that to work, `startHttpSpan` should keep the internal marker guard, but should not reject requests just because they already have `traceparent`. Otherwise `.deferToResume` still won’t produce a span.

carlosmcevilly (changes requested):

SplunkNetwork/Sources/SplunkNetwork/URLSessionInstrumentation/Instrumentation+TaskCreation.swift
SMickelsn: Thank you.   I've put in fixes addressing these issues.

carlosmcevilly (approved):
lgtm

PR #687, opened 2026-07-02
DEMRUM-5867: Batch trace spans in memory instead of exporting one per span

aditi-s3:
### Description
Replaces `SimpleSpanProcessor` with a new module-internal in-memory batching span processor (`OTLPBatchSpanProcessor`) on the **direct-span** trace pipeline. Ended spans are pooled in a bounded in-memory queue and handed to the existing disk-backed exporter in batches — every **0.5s** or when **100 spans** accumulate, whichever comes first — plus on app background/terminate/shutdown. This coalesces many small disk writes into fewer, larger ones.

**Changes**
- New `OTLPBatchSpanProcessor` + `BatchSpanProcessorCore` (split across `OTLPBatchSpanProcessor.swift` and `BatchSpanProcessorCore+Draining.swift` for lint length limits; everything is module-internal, no new public API).
- New `ReadableSpanQueue`: fixed-capacity FIFO ring buffer (default 2048); newest spans dropped when full.
- `OTLPTraceProcessor` swapped from `SimpleSpanProcessor` to `OTLPBatchSpanProcessor`.
- `SpanInterceptorExporter` and `SplunkStdoutSpanExporter` now forward `explicitTimeout` (previously ignored) so bounded operations stay bounded end-to-end.
- Docs: `AGENTS.md` Telemetry Model + `CHANGELOG.md`.
- Tests: `OTLPBatchSpanProcessorTests`, `OTLPBatchProcessorSafetyTests`, `ReadableSpanQueueTests`, plus test-support exporters.
Session Replay and the log-as-span pipeline are **unchanged**; only the direct-span processor is swapped.

**Context: design choices & deliberate trade-offs (please review)**
- **Crash-window loss (central trade-off):** spans buffered in memory when the app crashes or is force-killed are **lost by design**, in exchange for fewer disk writes. Normal in-memory window is ≤0.5s or ≤100 spans. Durable buffering still happens downstream (background exporter is disk-backed); only the pre-disk window changes.
- **Host-app safety > telemetry completeness** wherever they conflict.
- **Never hang the main thread.** `runOnProcessorQueue` runs inline if already on `processorQueue` (avoids `sync` self-deadlock via `DispatchSpecificKey`), synchronous off-main, and on main:
  - `forceFlush`: **async, no wait** (non-terminal; undrained spans persist a moment later anyway).
  - `shutdown`: **bounded wait ≤ 1s** (one-shot teardown; no guaranteed later chance).
  - `willTerminate`: **bounded wait ≤ 2s** (process dying; cap under OS watchdog).
  - `didEnterBackground`: **async** (app stays alive; favor responsiveness).
- **Failure handling by intent:** background/`forceFlush`/periodic **requeue** failed batches (transient errors don't permanently lose spans); shutdown/terminate **drop** on failure to guarantee termination within the deadline.
- **Partial requeue:** `prependAsMuchAsPossible` puts back as much of a failed batch as fits; only true overflow is counted as dropped.
- **Bounded drains:** `drainSnapshot` drains only the count present at entry and checks the deadline between batches (concurrent producers can't starve the caller).
- **Coalesced flushes** (≤1 pending drain closure) and **clamped schedule delay** (floor 0.1s) prevent busy-looping.
- **Observable drops:** overflow + failed exports counted and logged (`.error`, throttled to ≤1 line / 5s with delta + running session total) rather than silently dropped.
- **Concurrency:** single serial `processorQueue` (`.utility`) owns the timer/exports; `NSLock` guards queue + counters, per repo concurrency guidance.

### Checklist

- [ ] I have performed a self-review of my code.
- [ ] All new and existing tests pass locally.
- [x] \(Optional) I have updated any relevant documentation.

### Generative AI usage

1. Temporarily enable the `SplunkOpenTelemetryTests` target in `SplunkAgent.xctestplan` (revert before committing — do not enable the flaky bundle in CI as part of this PR).

3. Run only the batching/queue suites (avoids the flaky tests in the rest of the bundle):
xcodebuild -scheme SplunkAgent -destination "OS=,name="
test -only-testing:SplunkOpenTelemetryTests/OTLPBatchSpanProcessorTests
-only-testing:SplunkOpenTelemetryTests/OTLPBatchProcessorSafetyTests
-only-testing:SplunkOpenTelemetryTests/ReadableSpanQueueTests

Expected: all batching/queue tests pass (21 tests as of this PR).

4. (Optional) Integration check in a sample app: generate navigation/network spans, confirm they still export to Splunk in batches, and background/terminate the app to confirm the buffer drains without a UI hang

- Minimize **crash-window** telemetry loss by adding disk-backed (write-ahead) buffering for the in-memory queue, so spans buffered at crash/force-kill/OOM time survive — a case not covered by the bounded drains on graceful background/terminate/shutdown. This re-introduces some per-span disk cost, so it should be weighed against the disk-churn reduction this PR provides (e.g. opt-in, or persist only under specific conditions).

- Comment regarding above.
Yes, it’s feasible, but I would not make it the default in this PR.

The current PR’s main win is that spans stay in memory briefly and are written to the existing disk-backed exporter in batches. A write-ahead buffer would protect the pre-batch crash window, but only by moving some disk work back onto the span path. If the WAL is truly crash-resilient, it has to persist before the span is considered buffered; if it writes asynchronously or coalesces writes, it reduces the loss window but does not eliminate it.

The hard parts are not just writing spans to disk:
- ReadableSpan is not a great durable format, so the WAL would likely need to serialize SpanData or an OTLP-ready representation.
- On successful export, the WAL needs ack/compaction logic.
- If the process crashes after spanExporter.export persists the batch but before the WAL ack is recorded, replay can duplicate spans on next launch.
- If the WAL acks too early, it can lose spans.
- Corruption recovery, disk caps, schema versioning, cleanup, and startup replay all need careful tests.
- Durable per-span writes, especially with fsync-style guarantees, are exactly the kind of host-app overhead this PR is trying to avoid.

My recommendation: do not add write-ahead buffering as part of this batching PR unless reviewers explicitly require it. I’d document the known trade-off instead.

If we do pursue it, I’d make it a separate, opt-in feature, something like “crash-resilient span buffering,” with a bounded append-only WAL and measurement. The default should stay memory batching + existing disk-backed exporter. A lighter middle ground would be best-effort async WAL or periodic snapshots, but that should be described honestly as reducing, not eliminating, crash-window loss.


SMickelsn:
@codex

vaclav-halik-smartlook:
The new batching tests look valuable, but they currently appear to live in `SplunkOpenTelemetryTests`, which is not part of the active `SplunkAgent` test plan/scheme. That means this PR can pass CI without running the tests that cover the new processor behavior. Please wire this test target into the scheme/test plan, or move the tests to a target already executed by CI.

SMickelsn:
> The new batching tests look valuable, but they currently appear to live in `SplunkOpenTelemetryTests`, which is not part of the active `SplunkAgent` test plan/scheme. That means this PR can pass CI without running the tests that cover the new processor behavior. Please wire this test target into the scheme/test plan, or move the tests to a target already executed by CI.

Updated: SplunkAgent.xctestplan includes SplunkOpenTelemetryTests with the batching suites selected:
OTLPBatchProcessorSafetyTests, OTLPBatchSpanProcessorLifecycleTests, OTLPBatchSpanProcessorTests, and ReadableSpanQueueTests.

SMickelsn:
Code moved to https://github.com/signalfx/splunk-otel-ios/pull/695

PR #695, opened 2026-07-14
DEMRUM-5867: Batching spans in memory

SMickelsn:
### Description
Direct spans are buffered in a bounded in-memory batch before being persisted by the disk-backed exporter. Background and normal termination notifications trigger asynchronous, best-effort drains without blocking the lifecycle notification thread. Spans may still be lost if the process exits before a drain completes.

### Initial Discussion
Replaces `SimpleSpanProcessor` with a new module-internal in-memory batching span processor (`OTLPBatchSpanProcessor`) on the **direct-span** trace pipeline. Ended spans are pooled in a bounded in-memory queue and handed to the existing disk-backed exporter in batches — every **0.5s** or when **100 spans** accumulate, whichever comes first — plus on app background/terminate/shutdown. This coalesces many small disk writes into fewer, larger ones.

**Changes**
- New `OTLPBatchSpanProcessor` + `BatchSpanProcessorCore` (split across `OTLPBatchSpanProcessor.swift` and `BatchSpanProcessorCore+Draining.swift` for lint length limits; everything is module-internal, no new public API).
- New `ReadableSpanQueue`: fixed-capacity FIFO ring buffer (default 2048); newest spans dropped when full.
- `OTLPTraceProcessor` swapped from `SimpleSpanProcessor` to `OTLPBatchSpanProcessor`.
- `SpanInterceptorExporter` and `SplunkStdoutSpanExporter` now forward `explicitTimeout` (previously ignored) so bounded operations stay bounded end-to-end.
- Docs: `AGENTS.md` Telemetry Model + `CHANGELOG.md`.
- Tests: `OTLPBatchSpanProcessorTests`, `OTLPBatchProcessorSafetyTests`, `ReadableSpanQueueTests`, plus test-support exporters.
Session Replay and the log-as-span pipeline are **unchanged**; only the direct-span processor is swapped.

**Context: design choices & deliberate trade-offs (please review)**
- **Crash-window loss (central trade-off):** spans buffered in memory when the app crashes or is force-killed are **lost by design**, in exchange for fewer disk writes. Normal in-memory window is ≤0.5s or ≤100 spans. Durable buffering still happens downstream (background exporter is disk-backed); only the pre-disk window changes.
- **Host-app safety > telemetry completeness** wherever they conflict.
- **Never hang the main thread.** `runOnProcessorQueue` runs inline if already on `processorQueue` (avoids `sync` self-deadlock via `DispatchSpecificKey`), synchronous off-main, and on main:
  - `forceFlush`: **async, no wait** (non-terminal; undrained spans persist a moment later anyway).
  - `shutdown`: **bounded wait ≤ 1s** (one-shot teardown; no guaranteed later chance).
  - `willTerminate`: **bounded wait ≤ 2s** (process dying; cap under OS watchdog).
  - `didEnterBackground`: **async** (app stays alive; favor responsiveness).
- **Failure handling by intent:** background/`forceFlush`/periodic **requeue** failed batches (transient errors don't permanently lose spans); shutdown/terminate **drop** on failure to guarantee termination within the deadline.
- **Partial requeue:** `prependAsMuchAsPossible` puts back as much of a failed batch as fits; only true overflow is counted as dropped.
- **Bounded drains:** `drainSnapshot` drains only the count present at entry and checks the deadline between batches (concurrent producers can't starve the caller).
- **Coalesced flushes** (≤1 pending drain closure) and **clamped schedule delay** (floor 0.1s) prevent busy-looping.
- **Observable drops:** overflow + failed exports counted and logged (`.error`, throttled to ≤1 line / 5s with delta + running session total) rather than silently dropped.
- **Concurrency:** single serial `processorQueue` (`.utility`) owns the timer/exports; `NSLock` guards queue + counters, per repo concurrency guidance.

### Checklist

- [x] \(Optional) I have updated any relevant documentation.

### Generative AI usage

1. Temporarily enable the `SplunkOpenTelemetryTests` target in `SplunkAgent.xctestplan` (revert before committing — do not enable the flaky bundle in CI as part of this PR).

3. Run only the batching/queue suites (avoids the flaky tests in the rest of the bundle):
xcodebuild -scheme SplunkAgent -destination "OS=,name="
test -only-testing:SplunkOpenTelemetryTests/OTLPBatchSpanProcessorTests
-only-testing:SplunkOpenTelemetryTests/OTLPBatchProcessorSafetyTests
-only-testing:SplunkOpenTelemetryTests/ReadableSpanQueueTests

Expected: all batching/queue tests pass (21 tests as of this PR).

4. (Optional) Integration check in a sample app: generate navigation/network spans, confirm they still export to Splunk in batches, and background/terminate the app to confirm the buffer drains without a UI hang

- Minimize **crash-window** telemetry loss by adding disk-backed (write-ahead) buffering for the in-memory queue, so spans buffered at crash/force-kill/OOM time survive — a case not covered by the bounded drains on graceful background/terminate/shutdown. This re-introduces some per-span disk cost, so it should be weighed against the disk-churn reduction this PR provides (e.g. opt-in, or persist only under specific conditions).

- Comment regarding above.
Yes, it’s feasible, but I would not make it the default in this PR.

The current PR’s main win is that spans stay in memory briefly and are written to the existing disk-backed exporter in batches. A write-ahead buffer would protect the pre-batch crash window, but only by moving some disk work back onto the span path. If the WAL is truly crash-resilient, it has to persist before the span is considered buffered; if it writes asynchronously or coalesces writes, it reduces the loss window but does not eliminate it.

The hard parts are not just writing spans to disk:
- ReadableSpan is not a great durable format, so the WAL would likely need to serialize SpanData or an OTLP-ready representation.
- On successful export, the WAL needs ack/compaction logic.
- If the process crashes after spanExporter.export persists the batch but before the WAL ack is recorded, replay can duplicate spans on next launch.
- If the WAL acks too early, it can lose spans.
- Corruption recovery, disk caps, schema versioning, cleanup, and startup replay all need careful tests.
- Durable per-span writes, especially with fsync-style guarantees, are exactly the kind of host-app overhead this PR is trying to avoid.

My recommendation: do not add write-ahead buffering as part of this batching PR unless reviewers explicitly require it. I’d document the known trade-off instead.

If we do pursue it, I’d make it a separate, opt-in feature, something like “crash-resilient span buffering,” with a bounded append-only WAL and measurement. The default should stay memory batching + existing disk-backed exporter. A lighter middle ground would be best-effort async WAL or periodic snapshots, but that should be described honestly as reducing, not eliminating, crash-window loss.


SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Trace Processor/OTLPTraceProcessor.swift
```diff
@@ -106,6 +134,34 @@ public class OTLPTraceProcessor: TraceProcessor {
...
+            )
+            .add(spanProcessor: OTLPImmediateSpanProcessor(spanExporter: spanInterceptorExporter))
+            .build()
+    }
+
+
+    // MARK: - Lifecycle
+
+    /// Registers the direct-span termination drain.
+    public func registerTerminationObserver() {
```
aditi-s3: @SMickelsn Why is this a public hook?

SplunkSlowFrameDetector/Sources/SplunkSlowFrameDetector/SlowFrameDetector.swift
```diff
@@ -193,6 +193,10 @@ public final class SlowFrameDetector: NSObject {
         await logic.flushBuffers()
     }
 
+    package func flushBufferedTelemetryForTermination() async {
```
aditi-s3: @SMickelsn Should you check if the module is enabled first?

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Span Processors/OTLPImmediateSpanProcessor.swift
aditi-s3, line 1: Do we really need the OTLPImmediateSpanProcessor ? Can both direct span and logs-to-span signals use one OTLPBatchSpanProcessor? Why are the drawbacks

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Trace Processor/OTLPTraceProcessor.swift
```diff
@@ -32,12 +32,18 @@ public class OTLPTraceProcessor: TraceProcessor {
     private let tracerProvider: TracerProvider
 
     /// Background exporter for traces (stored to support endpoint updates).
-    private let backgroundTraceExporter: OTLPBackgroundHTTPTraceExporter
+    private let backgroundTraceExporter: any EndpointConfigurableSpanExporter
+
+    /// In-memory batch processor for direct spans.
+    private let batchSpanProcessor: OTLPBatchSpanProcessor
+
+    /// Tracer provider for log-to-span telemetry that must bypass the in-memory span batch.
```
aditi-s3: @SMickelsn Why do we want to bypass the batching for logs that are ultimately exported as spans?

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Trace Processor/OTLPTraceProcessor.swift
SMickelsn: Good catch. There is no customer-facing use case for this hook.  Fixing

SplunkSlowFrameDetector/Sources/SplunkSlowFrameDetector/SlowFrameDetector.swift
SMickelsn: Yes, agreed. A disabled detector currently has no buffered telemetry, so this is effectively a no-op, but checking isEnabled avoids unnecessary termination work. I’ll add the guard.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Span Processors/OTLPImmediateSpanProcessor.swift
SMickelsn: I have an elaborate reply from codex, pasted below

With a shared OTLPBatchSpanProcessor, ending the converted span would only enqueue it in memory. The completion would run before persistence.

Drawbacks of sharing one batch processor:
- Log-to-span events could be lost if the app crashes or is force-killed before the 0.5-second batch drain.
- Crash, error, and internal-event telemetry would compete with direct spans for the 2,048-span queue.
- A full queue could drop higher-value crash/error telemetry.
- Existing completion semantics would change from “reached the disk-backed exporter” to “accepted into memory.”
- Termination would need to drain the log-processing queue before snapshotting the span batch; the current termination drain does not provide that coordination.

Drawbacks of the immediate processor:
- One JSON encoding and disk write per log-derived span.
- More disk files and I/O than batching.
- A second tracer provider and additional pipeline complexity.
- Synchronous work can stall the caller; most log events run on a utility queue, but immediate paths must be handled carefully.

We need a separate processing policy if we want to retain the existing log-to-span durability and completion semantics. With the batch processor, emit() would return after the converted span was only queued in memory, so crash/error events could be lost before the next drain and would compete with direct spans for queue capacity. The immediate processor ensures the converted span reaches the disk-backed exporter before completion. The tradeoff is per-event encoding/disk I/O and an additional provider. If we are willing to redefine completion as “accepted into memory” and accept the same crash-loss window as direct spans, then both paths could share the batch processor.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Trace Processor/OTLPTraceProcessor.swift
SMickelsn: I think this is essentially a duplicate of the above issue, and would be satisfied by the resolution of that one.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Span Processors/OTLPImmediateSpanProcessor.swift
aditi-s3: @SMickelsn 
I think both direct and converted spans can accept the same batching tradeoff of loss of telemetry.
The current production SimpleSpanProcessor exports asynchronously, so log-event completion has not historically guaranteed disk persistence.
I would prefer one batch processor and a uniform policy for ordinary direct and log-derived spans. 
Crash reports are different because PLCrashReporter already keeps the source report on disk. So maybe there is a better way to avoid loss of crashes by retaining the PLCrashReporter report until the downstream exporter confirms disk persistence ? Just an idea

Please also get feedback from @vaclav-halik-smartlook and @carlosmcevilly on this

SMickelsn: I'm on it.   Have to give the PLCrash issue some thought.  I'd be concerned about tying export between the two modules and creating an elaborate api.   The holding until disk persistence will work.

vaclav-halik-smartlook: **I’d also like to add a few of my own thoughts on this topic:**
- The whole system is already complex, and I’d prefer to **keep it as simple as possible**.
- There will always be data we want to “save” because it’s valuable (e.g., crash reports). The problem is determining what is valuable enough to require special care. For some, it’s just the report, but for another user, it might be the entire path leading up to the crash, including interactions or network communication.
- That brings me back to the original problem. By its nature, OTel produces many small signals; logging them is expensive, and in the event of a crash, it’s difficult to troubleshoot.

**In my opinion, there’s only one reasonably feasible approach here:**
Accept that in a crash, we’ll lose much essential data. Handle crash reports within the module using a special, secure method, ideally by restoring the saved report on the next startup. However, this cannot be used in other modules because it requires low-level APIs (which the current PLCrashReporter already does).

I would base this on the assumption that, while we may not have all the most recent signals (such as the button the user clicked or an error during decoding of an incoming response), we do have the crash report and a significant portion of the path the user took to reach that point. That should be sufficient to resolve any potential issue.

We also need to take into account that app crashes are only a subset of what clients actually deal with.

SplunkAgent/Sources/SplunkAgent/Agent/Modules/SplunkRum+Modules.swift
```diff
@@ -72,6 +72,24 @@ extension SplunkRum {
         customizeSlowFrameDetector()
     }
 
+    func configureTraceLifecycleDrains() {
```
vaclav-halik-smartlook: **issue:** This is in the wrong place and lacks a clear structure. If any module-specific configuration is needed, it should be included in the relevant methods that are already here. Please do not create additional procedures.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Span Processors/ReadableSpanQueue.swift
```diff
@@ -0,0 +1,105 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import OpenTelemetrySdk
+
+/// Fixed-capacity FIFO storage that avoids shifting retained spans while producers hold the state lock.
+struct ReadableSpanQueue {
```
vaclav-halik-smartlook, line 30: **chore:** This class could have better documentation. It lacks a general description of what it achieves and why a regular array isn't used. Also, there's no description of the methods at all. I would add it; it's quite useful and aligns with our goal of building reusable components in the future.

SplunkOpenTelemetryBackgroundExporter/Sources/SplunkOpenTelemetryBackgroundExporter/OTLPBackgroundHTTPBaseExporter.swift
```diff
@@ -277,9 +310,8 @@ extension OTLPBackgroundHTTPBaseExporter {
     }
 
     /// Flushes any data stored in pending storage by moving it to active storage and scheduling uploads.
-    private func flushPendingData() {
-        let (currentEndpoint, currentHeaders) = getEndpointState()
-        guard let currentEndpoint else {
+    private func flushPendingData(for activationRevision: UInt64) {
```
vaclav-halik-smartlook, line 328: [P2] Pending payload can get stuck in active storage without being scheduled

`flushPendingData(for:)` copies a pending file into active storage and deletes the pending copy before it re-checks whether the endpoint revision is still current. If `clearEndpoint()` or another `setEndpoint()` happens between the delete and the later `endpointState(for:)` check, the method returns without calling `httpClient.send`.

At that point the payload is no longer in pending storage, so the next endpoint activation will not pick it up through `flushPendingData`; it only depends on the delayed stalled-upload check to discover and schedule the active file later. That creates an avoidable delay, and if stalled-upload recovery is disabled or does not run, the payload can remain unscheduled.

Please keep the file in pending storage until the activation is still valid and scheduling is committed, or explicitly move it back / schedule active files immediately when activation is canceled.

SplunkAgent/Sources/SplunkAgent/Agent/Events/Event Manager/DefaultEventManager+Endpoint.swift
```diff
@@ -22,6 +22,16 @@ internal import SplunkOpenTelemetry
...
 
+    /// Registers the direct-span background drain that runs after async producers flush their spans.
+    func registerTraceBackgroundDrain(prepareForBackground: @escaping () async -> Void) {
+        concreteTraceProcessor.registerBackgroundObserver(prepareForBackground: prepareForBackground)
+    }
+
+    /// Registers the direct-span terminal drain that runs after terminal producers flush their spans.
+    func registerTraceTerminationDrain(prepareForTermination: @escaping () async -> Void) {
+        concreteTraceProcessor.registerTerminationObserver(prepareForTermination: prepareForTermination)
+    }
```
vaclav-halik-smartlook: **question:** I’m not fully convinced this lifecycle drain registration pays for its added coupling/complexity. It assumes we can maintain a complete manual list of async span producers that must run before the batch snapshot, but today this is wired only from `SplunkRum` for specific modules. Future producers can easily be missed, and the `register...` naming suggests multiple registrations even though the underlying observer is effectively single-registration.

Can we either document the concrete producer problem this solves, or simplify this by keeping the batch processor lifecycle drain independent of module-specific prepare hooks? For background the normal batch timer already limits the extra in-memory window to ~0.5s, and for termination this remains best-effort anyway.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Log Event Processors/OTLPLogToSpanEventProcessor.swift
```diff
@@ -49,6 +49,40 @@ public class OTLPLogToSpanEventProcessor: LogEventProcessor {
...
+            debugEnabled: debugEnabled,
+            tracerProvider: traceProcessor.tracerProvider
+        )
+    }
+
+    init(
+        with _: URL?,
+        resources: AgentResources,
+        debugEnabled: Bool,
+        tracerProvider: any TracerProvider
```
vaclav-halik-smartlook: **chore:** The tracer-provider injection makes sense here, but the initializer structure feels more complicated than it needs to be. Right now there are multiple initialization paths that repeat the same resource setup and logger-provider construction, which makes future changes easier to apply inconsistently.

Could we keep a single designated initializer that takes a `tracerProviderProvider` closure, and have the existing public initializer plus the trace-processor/test overloads delegate to it? That would preserve the current API shape while keeping all setup logic in one place.

vaclav-halik-smartlook (changes requested):
I haven't done a review of the changes to the `Span Processors` yet, because I'm not sure if this design will end up being scrapped. However, I can do it as a second round.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Span Processors/ReadableSpanQueue.swift
SMickelsn: Agreed, and thanks

SplunkOpenTelemetryBackgroundExporter/Sources/SplunkOpenTelemetryBackgroundExporter/OTLPBackgroundHTTPBaseExporter.swift
SMickelsn: Good catch

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Log Event Processors/OTLPLogToSpanEventProcessor.swift
SMickelsn: The consolidation goal is good, but the exact proposal has a compatibility concern: the existing public required init is a designated initializer. Making it delegate to a new closure-based initializer would require changing it to convenience, preventing external subclass designated initializers from calling it via super.init.

The safer refactor is:
- Keep the existing public required initializer designated.
- Add one private initialization-state factory accepting tracerProviderProvider.
- Have both designated initializers use that factory.
- Keep the trace-processor convenience initializer delegating to the injected-provider initializer.

That removes duplicated resource and logger-provider construction while preserving the public initializer’s designated status and subclassing behavior.

Repaired.

SplunkAgent/Sources/SplunkAgent/Agent/Events/Event Manager/DefaultEventManager+Endpoint.swift
SMickelsn: The hook addresses a concrete race in SlowFrameDetector: lifecycle notifications are delivered through an AsyncStream and actor, so its final spans may be enqueued after the batch processor’s background or termination snapshot. The 0.5-second timer is not sufficient after backgrounding because the process can be suspended before it fires. I’ll document this narrow contract and clarify that lifecycle drain installation is single-registration. If additional async lifecycle producers appear, we should replace the explicit wiring with a common protocol/registry.

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Trace Processor/OTLPTraceProcessor.swift
```diff
@@ -26,18 +26,27 @@ import SplunkOpenTelemetryBackgroundExporter
...
     /// Background exporter for traces (stored to support endpoint updates).
-    private let backgroundTraceExporter: OTLPBackgroundHTTPTraceExporter
+    private let backgroundTraceExporter: any EndpointConfigurableSpanExporter
 
+    /// In-memory batch processor for direct spans.
+    private let batchSpanProcessor: OTLPBatchSpanProcessor
 
     // MARK: - Initialization
 
-    public required init(
```
aditi-s3, line 40: Remove convenience modifier. No longer needed

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Log Event Processors/OTLPLogToSpanEventProcessor.swift
```diff
@@ -50,17 +60,84 @@ public class OTLPLogToSpanEventProcessor: LogEventProcessor {
...
+        )
+
+        loggerProvider = state.loggerProvider
+
+        #if DEBUG
+            resource = state.resource
+        #endif
+    }
+
+    public convenience init(
```
aditi-s3: @SMickelsn  This should not be public

SplunkAgent/Sources/SplunkAgent/Agent/Events/Event Manager/DefaultEventManager+Endpoint.swift
aditi-s3: @SMickelsn This brings added cross-module complexity and maintenance cost though. 

I would consider removing configureTraceLifecycleDrains() and its forwarding methods, along with the registration methods and changes it requires. This avoids coupling the generic trace processor to individual instrumentation modules.
If we want a best-effort background drain, it can remain entirely inside the batch processor:

```
private func installBackgroundDrain() {
    backgroundObserver = NotificationCenter.default.addObserver(
        forName: UIApplication.didEnterBackgroundNotification,
        object: nil,
        queue: nil
    ) { [weak self] _ in
        self?.processorQueue.async {
            self?.drainSnapshot(deadline: nil, requeueOnFailure: true)
        }
    }
}
```
and live with the tradeoff asynchronous background drain. We could also run a live test to see the material loss but maybe consider it as a future improvement after some testing. I will leave it up to your discretion to see if this is truly valuable

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Trace Processor/OTLPTraceProcessor.swift
SMickelsn: Done

SplunkOpenTelemetry/Sources/SplunkOpenTelemetry/Log Event Processors/OTLPLogToSpanEventProcessor.swift
SMickelsn: There have been a number of changes that I think addresses issue

carlosmcevilly (commented):
LGTM. I've been going at this one intensely with Codex but you've also been updating at the same time, and every code issue it flagged has been fixed now.

For my other thoughts on the PR shared in chat I agree that this is a good step for now, and great work! And we can consider other use cases as we go forward, in separate PRs.

The only remaining things flagged by Codex are some staleness in the PR description now that some new changes have been introduced. I'll let you make the call on that.

And similarly some staleness in part of AGENTS.md:

Codex:

"Export buffering is disk-backed, not an in-memory bounded queue" under Design Assumptions is now directly incorrect. The accurate model is:

`span creation → bounded in-memory batch → disk-backed exporter → background upload`

- Its lifecycle wording should ideally say background and termination drains are best effort and asynchronous, rather than implying a guaranteed drain.

aditi-s3 (approved):

carlosmcevilly (approved):

vaclav-halik-smartlook (approved):
After incorporating the changes, it looks good.

